PR #18, opened 2020-10-27
Go modules

hugelgupf:
Do not merge. I may need to break u-root first.


rminnich:
I got some changes from here, updated our requirements to go1.17, installed cox via apt install instead of go get, and ... it passed. 

Modules get a little less hard each iteration of go ... cross fingers.

PR #62, opened 2022-03-03
cpud: add -fstab switch

rminnich:
CPU_FSTAB is a string in fstab(5) format, which allows arbitrary
fstab usage. For multiple mounts, just embed newlines in the string
or see the example below for how to populate it from a file.

If we have a guest VM, we can now use virtio-fs and the performance
improvement is pretty stunning.

You can do this:
CPU_NAMESPACE="" PWD=/ CPU_FSTAB="myfs /mnt virtiofs rw 0 0" cpu v ls /mnt

or
CPU_NAMESPACE="" PWD=/ CPU_FSTAB=${fstab} cpu v ls /mnt

for more complex mounts, if you can figure out how to populate an
environment variable from a file without having newlines converted to
spaces :-)

Current limits: code does not parse the options string, so you can not say
CPU_NAMESPACE="" PWD=/ CPU_FSTAB="myfs /mnt virtiofs defaults 0 0" cpu v ls /mnt

since useful defaults are file-system-type specific.

### DIFF
--- a/cmds/cpu/cpu.go
+++ b/cmds/cpu/cpu.go
@@ -45,6 +45,7 @@ var (
 	debug       = flag.Bool("d", false, "enable debug prints")
 	dbg9p       = flag.Bool("dbg9p", false, "show 9p io")
 	dump        = flag.Bool("dump", false, "Dump copious output, including a 9p trace, to a temp file at exit")
+	fstab       = flag.String("fstab", "", "pass an fstab to the cpud")
 	hostKeyFile = flag.String("hk", "" /*"/etc/ssh/ssh_host_rsa_key"*/, "file for host key")
 	keyFile     = flag.String("key", "", "key file")
 	mountopts   = flag.String("mountopts", "", "Extra options to add to the 9p mount")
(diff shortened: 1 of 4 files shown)


rminnich:
don't approve yet, we're going to make this a switch: 
-fstab file
longer FSTAB are hard to get into the command line.
-fstab file will place 
CPU_FSTAB 
in the environment for the cpud to use

rminnich:
ok good to go

cmds/cpu/cpu.go
```diff
@@ -203,16 +204,26 @@ func runClient(host, a, port, key string) error {
...
-func env(s *ossh.Session, envs ...string) {
+func env(s *ossh.Session, envs ...string) error {
 	for _, v := range append(os.Environ(), envs...) {
 		env := strings.SplitN(v, "=", 2)
 		if len(env) == 1 {
 			env = append(env, "")
 		}
 		if err := s.Setenv(env[0], env[1]); err != nil {
-			log.Printf("Warning: s.Setenv(%q, %q): %v", v, os.Getenv(v), err)
+			return fmt.Errorf("Warning: s.Setenv(%q, %q): %v", v, os.Getenv(v), err)
```
orangecms, line 214: Should this change be reverted? It looks like a leftover from the env var approach.

orangecms: Huh, now I get it... the env var is abused for marshalling. Was a bit unintuitive first. 😅

orangecms (approved):

cmds/cpu/cpu.go
rminnich: This also passes all environment variables.

rminnich: hence os.Environ()

PR #69, opened 2022-03-08
cpu: handle "os: invalid use of WriteAt on file opened with O_APPEND"

### LINKED ISSUE #64: invalid use of WriteAt on file opened with O_APPEND
After running a command with bash or sh and hitting ctl-d or type 'exit', I get that error.  If i hit ctl-d immediately, i do not.  if the command is 'exit', i also get it.

seems like any time i issue any command, even 'exit', i'll get the error.  but not if if i hit ctl-d first.  (and not if i get some other error, as in the other bug, where ctl-c followed by ctl-d gave a different error).

e.g.
```
$ ./cpu TARGET sh
sh-5.1# ls
cpu	  cpu_darwin.go  cpu_linux.go  getattr_darwin.go  srv.go
cpu9p.go  cpu.go	 doc.go        getattr_linux.go
sh-5.1# exit
exit
2022/03/04 16:27:08 unknown error: os: invalid use of WriteAt on file opened with O_APPEND
                                                                                          $ 

$ ./cpu TARGET sh
sh-5.1# exit
exit
2022/03/04 16:27:32 unknown error: os: invalid use of WriteAt on file opened with O_APPEND
                                                                                           $ 

(i hit ctl-d on this one)
$ ./cpu TARGET sh
sh-5.1# 
exit

```

rminnich:
The 9p Twrite message is equivalent to WriteAt, i.e. it takes data and offset.

Files can be opened as O_APPEND. On Plan 9, all servers use O_APPEND to indicate
"ignore offset, just write at end." 9P Twrite and Plan 9 system call
are thus self-consistent: servers ignore offset on files opened O_APPEND,
honor offset otherwise; clients open files with or without O_APPEND.

Linux, other Unix-like systems, and Windows do not seem to share this
consistency. Mixing write at offset and O_APPEND writes does not work
at all, and will corrupt the file.

In c7f7f5, in March 2019, the Go runtime changed to not allow mixed
pwrite and append writes.

This is likely because Linux and other sytems allow pwrite on files
opened with O_APPEND, which will make a large mess of things. The author of that
commit comments:

    Making WriteAt returns an error if file is opened in append mode, we
    guarantee to get consistent behavior between platforms, also prevent
    user from accidently corrupting their data.

Hence the Go runtime, since 2019, returns an error for something that used
to work, even though on, e.g., Plan 9, there is no problem. This change
will likely not work for Plan 9 users, but few people care :-)

Fortunately, this change is easy to work around, and cheap. It rarely
occurs; it is generated in the Go runtime by a test of a (non exported)
member of a struct; no system calls are involved.

Hence, we need only call WriteAt and, in the rare event of an error,
see if err.Error() contains the string (the error value is not exported,
unfortunately). If it does contain the string, do a Write and return
what Write returns.

Fixes #64.


cmds/cpu/cpu9p.go
```diff
@@ -144,8 +145,21 @@ func (l *cpu9p) ReadAt(p []byte, offset int64) (int, error) {
 }
 
 // Write implements p9.File.WriteAt.
+// There is a very rare case where O_APPEND files are written more than
```
brho, line 148: what's interesting is Go's WriteAt fails anytime O_APPEND was set, yet we're able to get a single write through.

my guess is that Create wasn't passing the O_APPEND along, such that if you created a file with O_APPEND, the Go runtime wouldn't have f.appendmode set, so WriteAt would work. 

from looking at os file.go, there's a Create that makes up its own O_flags, and O_APPEND isn't one of them.  so i guess if you wanted to do O_CREATE | O_APPEND, you'd have to use OpenFile or whatever.  arguably, no one really wants to do that, and the only reason the shell does it is that it's either CREATE or APPEND.  

though there is a subtle difference between a freshly created file without append, and a freshly created file with append: for the latter, the OS will automatically seek your file for you to the end of the inode.  so if you have two threads writing to a log file, you'd get the corruption that presumably Go is trying to avoid.  i.e. you'd have to do OpenFile(O_CREATE | O_APPEND) so that the creator is also in append mode.

anyway, this is subtle shit, and mildly annoying that you have to work around the runtime.  imagine if glibc wouldn't let you do a pwrite() to an O_APPEND fd!

in go's defense, posix and linux disagree on the behavior of pwrite and O_APPEND:
```
BUGS
       POSIX requires that opening a file with the O_APPEND flag should have no effect on the location at which pwrite() writes data.  However, on Linux, if  a
       file is opened with O_APPEND, pwrite() appends data to the end of the file, regardless of the value of offset.

```

cmds/cpu/cpu9p.go
```diff
@@ -144,8 +145,21 @@ func (l *cpu9p) ReadAt(p []byte, offset int64) (int, error) {
...
+// after checking the open flag in the os.File struct.
+// I.e. the error is not generated by a system call,
+// so it is very cheap to try the WriteAt, check the
+// error, and call Write if it is the rare case of a second write
+// to an append-only file..
 func (l *cpu9p) WriteAt(p []byte, offset int64) (int, error) {
-	return l.file.WriteAt(p, int64(offset))
+	n, err := l.file.WriteAt(p, int64(offset))
+	if err != nil {
+		if strings.Contains(err.Error(), "os: invalid use of WriteAt on file opened with O_APPEND") {
```
brho, line 158: it's a very Plan9 thing to treat error strings as ABI.  =)

does Go treat error strings as a stable ABI?  if not, if we ever have easy tests for cpu, we can make sure the O_APPEND case is tested so we can catch the breakage.

brho (approved):

cmds/cpu/cpu9p.go
rminnich: yep, it's a problem. Few people know or care however :-(

PR #71, opened 2022-03-08
rust: try 2

rminnich:
I did not find thrussh particulary nice to use.


orangecms:
Gotcha... I feel like it's a bit of OOesque/Trait vs imperative/functional style. I also prefer the latter, though I'm okay with both. thrussh has quite some issues open though, no idea about this one here. I'll keep the thrussh thing around so that we can compare again when the code grows. 🌱

orangecms (approved):

PR #72, opened 2022-03-09
Add a -namespace switch to the cpu command

rminnich:
Until now, there has been a default namespace in the server.
/bin:/lib:/lib64:/usr/lib:/home

The problem is that, in the new world in which the namespace
can also come from an fstab, we've had no way to say that
the server need not mount /tmp/cpu; and the client need not
start a 9p server. Part of the problem is that embedding
a default namespace in the server was a foundational
mistake: the namespace, in all cases, should be specified
by the client.

To do so requires cpud parse CPU_NAMESPACE as follows:
o if CPU_NAMESPACE is not set, do not mount /tmp/cpu
o if it is set but empty, mount /tmp/cpu
o if it is not empty, do bind mounts specified

On the client, we might follow these rules too: just
blindly pass the environment, and if CPU_NAMESPACE is not
set, then the 9p server will not be started and the cpud
will not try to mount it.

This would fail the test of convenience: cpu should work
in a reasonable way with minimal options, and we should not
require users to know what to set.

To make all this work, add a -namespace switch to cpu,
with the reasonable value:
/bin:/lib:/lib64:/usr/lib:/home

cpu will continue to pass CPU_NAMESPACE if set. The
namespace switch, if it is not "none", will set and
possibly override CPU_NAMESPACE. The rules, then,
for namespace have three possibilities
o -namespace none and no CPU_NAMESPACE environment setting
   no 9p server started; cpud will not try to mount 9p
o -namespace default value (or set via -namespace)
  CPU_NAMESPACE sent to server will have value of namespace
o -namespace none and CPU_NAMESPACE has a value
  CPU_NAMESPACE sent to server and it will try to mount

This will make the creation of the rust client and server
a bit easier.


brho:
couple random thoughts:

- does it make sense for the user to put 9p mounts in fstab?  if i'm making my namespace from a few different mounts, that might be convenient.  if so, then maybe have another flag to control the client's 9p server?
- for rust, is the main helpful thing that there isn't a 9p server by default?
- if CPU_NAMESPACE is set and -namespace is set, then CPU_NAMESPACE is ignored?  or do we want an error/warning?

orangecms (approved):
Looks sensible.

The many go.{mod,sum} changes in this and other PRs sometimes are weird. Wasn't there a go mod tidy just the other day? Anyway, shouldn't be an issue here.

PR #73, opened 2022-03-09
Package

rminnich:


orangecms:
What does pkg/package mean here? I am not seeing it. 🤔 

Is that about the `"github.com/u-root/cpu/cpu"` import?

rminnich:
> What does pkg/package mean here? I am not seeing it. 🤔
> 
> Is that about the `"github.com/u-root/cpu/cpu"` import?

The culture has changed. pkg is no longer the standard. Packages start at top level now.
https://github.com/golang-standards/project-layout/issues/10

and you're supposed to use internal/ where possible to hide packages you don't want people to see

for $MODULES$

but cpu/cpu is still a package :-)

rminnich:
anyway it's not ready, this will be a series, *almost* there but not quite

Issue #40, opened 2021-09-15
spurious error return of -1 at end of successful invocation


rminnich:
```
rminnich@a300:~/mainboards/pcengines/apu2$ cpu apu2 date
Tue Sep 14 12:02:21 AM PDT 2021
2021/09/15 14:18:33 SSH error Process exited with status 4294967295
````
Not sure why this happens, but it is happening on a pcengines/apu2

**Comment by brho** [comment]
might be similar to what i get.  i can get a return of 1 by doing a ctl-c before ctl-d when running bash.

This is with just hitting ctl-d, and i get no error:
```
$ ./cpu TARGET bash --norc --noprofile
bash-5.1# 
exit
```

This is with hitting ctl-c, then ctl-d
```
$ ./cpu TARGET bash --norc --noprofile
bash-5.1# ^C
bash-5.1# 
exit
2022/03/04 16:14:28 CPUD(as remote):exit status 130
2022/03/04 16:14:28 SSH error Process exited with status 1
                                                          $
```

**Comment by brho** [comment]
do you still get this one?  i still get my version of it, every time.  i can try debugging it too.

**Comment by brho** [comment]
ok for my specific thing, that's actually not a bug.  the shell actually returns 130, just as the output says.  without using cpu at all:

```
brho@gnomeregan ~/ $ sh
sh-5.1$ 
exit
brho@gnomeregan ~/ $ echo $?
0
brho@gnomeregan ~/ $ sh
sh-5.1$ ^C
sh-5.1$ 
exit
brho@gnomeregan ~/ $ echo $?
130
```

wtf is 130?  that's [SIGINT + 128](https://unix.stackexchange.com/questions/386836/why-is-doing-an-exit-130-is-not-the-same-as-dying-of-sigint)

why is that the return value of sh?  (which is bash btw, on my system).  because it looks like when bash exits, its exit status is the status of its last command.  that's certainly the behavior of [bash scripts](https://tldp.org/LDP/abs/html/exit-status.html#:~:text=After%20a%20script%20terminates%2C%20a,range%201%20%2D%20255%20on%20error.&text=%23!%2Fbin%2Fbash%20echo%20hello,returned%20because%20command%20executed%20successfully.), and it makes sense they do it for the shell process itself too.

e.g. 
```
brho@gnomeregan ~/ $ sh
sh-5.1$ ls /no-such/file
ls: cannot access '/no-such/file': No such file or directory
sh-5.1$ 
exit
[...]

PR #74, opened 2022-03-10
minor things

brho:
feel free to drop any or all of these.


cmds/cpud/main.go
```diff
@@ -390,13 +396,17 @@ func errval(err error) error {
...
+		a = append([]string{a[0], "-d"}, a[1:]...)
+	}
 	v("handler: cmd is %v", a)
 	cmd := exec.Command(a[0], a[1:]...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{Cloneflags: syscall.CLONE_NEWNS}
 	cmd.Env = append(cmd.Env, s.Environ()...)
 	ptyReq, winCh, isPty := s.Pty()
 	if isPty {
 		cmd.Env = append(cmd.Env, fmt.Sprintf("TERM=%s", ptyReq.Term))
+		cmd.Env = append(cmd.Env, "CPUD_IGNORE_CMD_ERROR=1")
```
rminnich, line 409: I gotta know what you hit that needed this but I'm good with it. 

I also realize ... damn ... the environment variables should have been CPUD_ all this time. Oh well.

rminnich: If you can, add the docs for this CPUD_ to doc.go?

rminnich (approved):

cmds/cpud/main.go
brho: i described the issue in https://github.com/u-root/cpu/issues/40, but it turns out i had a different issue than you.  

basically bash's return value is the return value of the last command bash ran, so if you did "cpu $HOST bash", and whatever you entered happened to not return 0, (e.g. ls /no/such/file), then cpu would loudly complain about the error code.

i added a blurb to doc.go too, which we can expand with other CPUD env variables, such as the fstab stuff (?).

i also squeezed in a commit that removed cmds/**cpud**/doc.go, which was an older copy of cmds/**cpu**/doc.go

PR #75, opened 2022-03-16
src/client: partially rewrite to more idiomatic Rust

orangecms:
This also shortens the code and reduces some seeming boilerplate.
Turns out that many expressions can be written in a shorter, simpler way. :-)

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -82,47 +82,30 @@ fn read_config(p: &Path) -> SshConfig {
 
 fn connect(host: &str, params: &HostParams) {
     // Resolve host
-    let host = match params.host_name.as_deref() {
-        Some(h) => h,
-        None => host,
+    let host = params.host_name.as_deref().unwrap_or(host);
+    let port = params.port.unwrap_or(22);
+
+    let formatted_host : String;
+    let host: &str = if host.contains(':') {
+        host
+    } else {
+        formatted_host = format!("{}:{}", host, port);
+        &*formatted_host
     };
-    let port = match params.port {
-        None => 22,
-        Some(p) => p,
-    };
-    let host = match host.contains(':') {
-        true => host.to_string(),
-        false => format!("{}:{}", host, port),
-    };
+
     println!("Connecting to host {}...", host);
-    let socket_addresses: Vec<SocketAddr> = match host.to_socket_addrs() {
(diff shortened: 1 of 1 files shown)


rminnich (approved):
This is great, the irony here is that this is sample code from the ssh package :-)

src/client/mod.rs
```diff
@@ -82,47 +82,30 @@ fn read_config(p: &Path) -> SshConfig {
...
-    };
-    let port = match params.port {
-        None => 22,
-        Some(p) => p,
-    };
-    let host = match host.contains(':') {
-        true => host.to_string(),
-        false => format!("{}:{}", host, port),
+    let host = params.host_name.as_deref().unwrap_or(host);
+    let port = params.port.unwrap_or(22);
```
rminnich: oh, oops, can you change this to 23.

cpu uses the old telnet port.

I wonder if I should not ask ICANN for it :-)

Or should we use the Plan 9 standard -- tcp=ncpu port=17010

?

src/client/mod.rs
```diff
@@ -82,47 +82,30 @@ fn read_config(p: &Path) -> SshConfig {
...
+
+    let formatted_host : String;
+    let host: &str = if host.contains(':') {
```
rminnich, line 89: is this backwards? Don't you want if !host.contains?

rminnich (changes requested):

src/client/mod.rs
orangecms: Either is fine by me. I like the 23, and we could always make this a parameter, CPU_PORT env var or whatever...
ICANN sounds fun 😹

src/client/mod.rs
orangecms: Yea this is Rust's notion of a ternary expression, a little back-and-forth reading
sorta `host.contains(':') ? host : ...` in other languages

rminnich: oh nvm for a second I thought it was a split but it's not. Got it.

src/client/mod.rs
rminnich: let's go with 23 for now. That matches the Go code.

orangecms: ok done

rminnich (approved):

PR #83, opened 2022-03-30
Add the session package

rminnich:
The session package is ues to manage a cpu session.

It is intended to be called by cpud after a connection has been made.

### DIFF
--- a/session/fns.go
+++ b/session/fns.go
@@ -1,0 +1,46 @@
+// Copyright 2018-2022 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package session
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ParseBinds parses a CPU_NAMESPACE-style string to a
+// slice of Bind structures.
+func ParseBinds(s string) ([]Bind, error) {
+	var b = []Bind{}
+	if len(s) == 0 {
+		return b, nil
+	}
+	binds := strings.Split(s, ":")
+	for i, bind := range binds {
+		if len(bind) == 0 {
+			return nil, fmt.Errorf("bind: element %d is zero length", i)
+		}
+		// If the value is local=remote, len(c) will be 2.
+		// The value might be some weird degenerate form such as
+		// =name or name=. Both are considered to be an error.
+		// The convention is to split on the first =. It is not up
+		// to this code to determine that more than one = is an error
+		// There is no rule that filenames can not contain an '='!
+		c := strings.SplitN(bind, "=", 2)
+		if len(c) == 2 {
+			l, r := c[0], c[1]
+			if len(r) == 0 {
+				return nil, fmt.Errorf("bind: element %d:name in %q: zero-length remote name", i, bind)
+			}
+			if len(l) == 0 {
+				return nil, fmt.Errorf("bind: element %d:name in %q: zero-length local name", i, bind)
+
+			}
+			b = append(b, Bind{Local: c[0], Remote: c[1]})
+			continue
+		}
+		b = append(b, Bind{Local: c[0], Remote: c[0]})
+	}
+	return b, nil
+}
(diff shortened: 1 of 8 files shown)


rminnich:
I'll be chumping these in, unless I get something by tomorrow, as this is still a very low-volume, low-interest project.

session/fns.go
```diff
@@ -0,0 +1,46 @@
...
+		// to this code to determine that more than one = is an error
+		// There is no rule that filenames can not contain an '='!
+		c := strings.SplitN(bind, "=", 2)
+		if len(c) == 2 {
+			l, r := c[0], c[1]
+			if len(r) == 0 {
+				return nil, fmt.Errorf("bind: element %d:name in %q: zero-length remote name", i, bind)
+			}
+			if len(l) == 0 {
+				return nil, fmt.Errorf("bind: element %d:name in %q: zero-length local name", i, bind)
```
orangecms, line 37: I find this a little unintuitive; I thought that it would be implied that omitting the second part would mean that local and remote be the same, i.e., `/home:/foo=/bar` be considered a shorthand for `/home=/home:/foo=/bar`.

session/session.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return unix.Setregid(-1, gid)
+}
+
+// Terminal sets up an interactive terminal.
+func (s *Session) Terminal() error {
+	// for some reason echo is not set.
+	t, err := termios.New()
+	if err != nil {
+		return fmt.Errorf("CPUD:can't get a termios; oh well; %v", err)
```
orangecms: To distinguish from the following message, this could be "CPUD: can't create...".

session/fns.go
rminnich: I'll llook at that, many thanks for reviews. Going out for a walk but will fix when i get back.

rminnich: but your review makes me happy I split the server up into server and session :-)

rminnich: ah, it is. 
This particular line is catching errors of this form
/home=
b/c if there is an =, the right hand side can't be empty.

session/session.go
rminnich: on yeah this is a screwup, will fix.

session/fns.go
orangecms: ah got it, line 43 is where the shorthand form ends up 👍

session/session_linux.go
```diff
@@ -0,0 +1,113 @@
...
+	"path/filepath"
+	"syscall"
+
+	"github.com/hashicorp/go-multierror"
+	"golang.org/x/sys/unix"
+)
+
+// Namespace assembles a NameSpace for this cpud, iff CPU_NONCE
+// is set and len(s.binds) > 0.
+// NOTE: this assumes we were started with Unshareflags set to CLONE_NEWNS.
```
orangecms: Can we assert this here? Or do we get an error somehow?
How will the user notice, what is "you will be sad" in terms of errors raised?

rminnich: it's very hard and probably impossible to test for. The tests currently will detect it, but there's NO WAY in linux to ensure you have an unshared name space that I have found. I've looked. This is more a warning to our future selves or users of this package. 

The main way you'd notice is if you and another user, sharing a mount, started stepping on each other. 

Note, however, that cpud does the right thing, by setting Unshareflags to CLONE_NEWNS, and I got a fix into the go runtime that makes sure that CLONE_NEWNS works correctly.

But it's not easy. Linux namespaces are a pretty terrible mess.

orangecms: Would you mind expanding the warning to our future selves to reflect this? I wouldn't have guessed. =)

rminnich: Is the new text any better?

orangecms: Sounds good, yea! Nit: `CLONE_NEWS` sould be `CLONE_NEWNS`.

rminnich: done, thanks.

rminnich: hope this works, would like to get this in and move on to the server package. Thanks for reviews :-)

orangecms (approved):

PR #84, opened 2022-04-07
cpud: add the server package

rminnich:
cpud is an sshd which allows a port forward, and sets up a bind
mount for /tmp/cpu/local to /.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1,0 +1,154 @@
+// Copyright 2018-2022 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package server
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+	"syscall"
+	"unsafe"
+
+	// We use this ssh because it implements port redirection.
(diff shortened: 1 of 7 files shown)


server/server.go
```diff
@@ -0,0 +1,154 @@
...
+		RequestHandlers: map[string]ssh.RequestHandler{
+			"tcpip-forward":        forwardHandler.HandleSSHRequest,
+			"cancel-tcpip-forward": forwardHandler.HandleSSHRequest,
+		},
+		Handler: handler,
+	}
+
+	if err := server.SetOption(ssh.HostKeyFile(hostKeyFile)); err != nil {
+		// We don't much care about this, what's the right thing to do here?
+		// just printint from this function is kind of icky.
```
orangecms, line 151: What's the issue with just printing a warning like `fmt.Printf("Could not load supplied host key: %s", err)`?

orangecms (approved):
Looks fine so far to my humble eyes. Only the tests seem a little lengthy as of now, but that's okay for they are end-to-end tests.

server/server.go
rminnich: random prints from random packages for warnings, at scale, make debugging in places like Google painful.

PR #87, opened 2022-04-14
Mountpackage

rminnich:


mount/mount.go
```diff
@@ -0,0 +1,87 @@
...
+		lineno++
+		l := s.Text()
+		if strings.HasPrefix(l, "#") {
+			continue
+		}
+		f := strings.Fields(l)
+		// fstab is historical, pretty free format.
+		// Users may have dropped a random fstab in and we need
+		// to be forgiving.
+		// The last two fields no longer have any meaning or use.
```
brho, line 51: but we require them, even if they have no meaning for us.  it's not particularly 'forgiving' as the comment suggests.  i'm fine with whatever - just want to make sure we want to reject fstab lines that don't have the trailing 0 0.

mount/mount.go
```diff
@@ -0,0 +1,87 @@
...
+
+// parse takes a mount options string and transforms
+// elements as needed (e.g. 'defaults') so the kernel
+// will accept it.
+func parse(m string) string {
+	var ret []string
+	for _, f := range strings.Split(strings.TrimSpace(m), ",") {
+		switch f {
+		case "defaults":
+			ret = append(ret, "rw", "suid", "dev", "exec", "auto", "nouser", "async")
```
brho, line 86: will it be a problem if the same option, e.g. rw, appears more than once in ret?  or do we want a map?

mount/mount.go
```diff
@@ -0,0 +1,87 @@
...
+		// If you set this flag to MS_PRIVATE, on some file systems, you'll get an EINVAL.
+		var flags uintptr
+		dev, where, fstype, opts := f[0], f[1], f[2], f[3]
+		// The man page implies that the Linux kernel handles flags of "defaults"
+		// we do no further manipulation of opts.
+		if e := os.MkdirAll(where, 0666); e != nil && !os.IsExist(e) {
+			err = multierror.Append(err, e)
+			continue
+		}
+		if e := m(dev, where, fstype, flags, opts); e != nil {
```
brho: i was expecting you to call parse(opts) here before passing it to mount.

brho: oh, i see it in the next commit.

brho (approved):
minor comment

mount/mount.go
rminnich: I will test now.
Last one rules.
 2561  sudo mount -t tmpfs -o rw,ro,rw none /tmp/x
 2562  mount
 2563  sudo umount /tmp/x
 2564  sudo mount -t tmpfs -o rw,ro,rw,ro none /tmp/x
 2565  mount

mount/mount.go
rminnich: we do reject them. We can fix this comment as needed later I guess.

PR #89, opened 2022-04-18
Add support for 9p SetAttr. Needed for uses at Google at least.

rminnich:

### DIFF
--- a/cmds/cpu/cpu9p_unix.go
+++ b/cmds/cpu/cpu9p_unix.go
@@ -1,0 +1,81 @@
+// Copyright 2022 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build !windows && !plan9
+// +build !windows,!plan9
+
+package main
+
+import (
+	"os"
+	"time"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/hugelgupf/p9/p9"
+	"golang.org/x/sys/unix"
+)
+
+// SetAttr implements p9.File.SetAttr.
+func (l *cpu9p) SetAttr(mask p9.SetAttrMask, attr p9.SetAttr) error {
+	var err error
+	// Any or ALL can be set.
+	// We started out with this:
+	// err = multierror.Append
+	// for each case. It seemed clean. But it returns a non-nil
+	// error, even if all arguments are nil. Interesting.
+	// Don't believe HashiCorp's docs about result.ErrorOrNil():
+	// they're just flat-out wrong; even their example doesn't work.
+	// A setattr could include things to set,
+	// and a permission value that makes setting those
+	// things impossible. Therefore, do these
+	// permission-y things last:
+	// Permissions
+	// GID
+	// UID
+	// Since changing, e.g., UID or GID might make
+	// changing permissions impossible.
+	//
+	// The test actually caught this ...
+
+	if mask.Size {
+		if e := unix.Truncate(l.path, int64(attr.Size)); e != nil {
+			err = multierror.Append(e)
+		}
+	}
+	if mask.ATime || mask.MTime {
+		atime, mtime := time.Now(), time.Now()
+		if mask.ATimeNotSystemTime {
+			atime = time.Unix(int64(attr.ATimeSeconds), int64(attr.ATimeNanoSeconds))
+		}
+		if mask.MTimeNotSystemTime {
+			mtime = time.Unix(int64(attr.MTimeSeconds), int64(attr.MTimeNanoSeconds))
+		}
+		if e := os.Chtimes(l.path, atime, mtime); e != nil {
+			err = multierror.Append(e)
+		}
+	}
+
+	if mask.CTime {
+		// The Linux client sets CTime. I did not even know that was allowed.
+		// Change the create time of a file?
+		// if e := errors.New("Can not set CTime on Unix"); e != nil { err = multierror.Append(e)}
+	}
+	if mask.Permissions {
+		if e := unix.Chmod(l.path, uint32(attr.Permissions)); e != nil {
+			err = multierror.Append(e)
+		}
+	}
+
+	if mask.GID {
+		if e := unix.Chown(l.path, -1, int(attr.GID)); e != nil {
+			err = multierror.Append(e)
+		}
+	}
+	if mask.UID {
+		if e := unix.Chown(l.path, int(attr.UID), -1); e != nil {
+			err = multierror.Append(e)
+		}
+	}
+	return err
+}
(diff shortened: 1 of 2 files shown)


cmds/cpu/cpu9p_unix.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// SetAttr implements p9.File.SetAttr.
+func (l *cpu9p) SetAttr(valid p9.SetAttrMask, attr p9.SetAttr) error {
+	var err error
+	// Any or ALL can be set.
+	// We started out with this:
+	// err = multierror.Append
+	// for each case. It seemed clean. But it returns a non-nil
+	// error, even if all arguments are nil. Interesting.
```
orangecms, line 26: Just a note: I'd expect the API for multierror to be that I'd have to check for its length > 0 instead of nil.

cmds/cpu/cpu9p_unix.go
```diff
@@ -0,0 +1,79 @@
...
+	// A setattr could include things to set,
+	// and a permission value that makes setting those
+	// things impossible. Therefore, do these
+	// permission-y things last:
+	// Permissions
+	// GUID
+	// UID
+	// Since changing, e.g., UID or GUID might make
```
orangecms: Nit: I know you love GUIDs, but this is just GID. :-)

cmds/cpu/cpu9p_unix.go
```diff
@@ -0,0 +1,79 @@
...
+	// We started out with this:
+	// err = multierror.Append
+	// for each case. It seemed clean. But it returns a non-nil
+	// error, even if all arguments are nil. Interesting.
+	// A setattr could include things to set,
+	// and a permission value that makes setting those
+	// things impossible. Therefore, do these
+	// permission-y things last:
+	// Permissions
+	// GUID
```
orangecms: just GID, no U

cmds/cpu/cpu9p_unix.go
```diff
@@ -0,0 +1,79 @@
...
+	"os"
+	"time"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/hugelgupf/p9/p9"
+	"golang.org/x/sys/unix"
+)
+
+// SetAttr implements p9.File.SetAttr.
+func (l *cpu9p) SetAttr(valid p9.SetAttrMask, attr p9.SetAttr) error {
```
orangecms: Why call it `valid`? Below, it sounds like it could as well not be. 🤔

orangecms: Suggestion: call it `mask`

cmds/cpu/cpu9p_unix_test.go
```diff
@@ -0,0 +1,199 @@
...
+	sam.MTime = true
+	if err := c.SetAttr(sam, sa); err != nil {
+		t.Fatalf("Setattr with MTime: %v != nil", err)
+	}
+	sam.ATimeNotSystemTime, sam.MTime = true, false
+	if err := c.SetAttr(sam, sa); err != nil {
+		t.Fatalf("Setattr with ATime, sam.ATimeNotSystemTime: %v != nil", err)
+	}
+
+	q, gm, ga2, err = c.GetAttr(m)
```
orangecms, line 154: this makes the test stateful and confusing - a simple table test would be clearer with one step at a time

orangecms: (or at least split into multiple parts, as some cases are deliberately making subsequent changes, being inherently stateful)

rminnich: I thought of a table but it's a bit messier than it seems at first.

cmds/cpu/cpu9p_unix.go
rminnich: multierror is not an array, so it would not have length. But it does have result.ErrorOrNil() so i'll use that.

cmds/cpu/cpu9p_unix.go
rminnich: done

cmds/cpu/cpu9p_unix.go
rminnich: done

cmds/cpu/cpu9p_unix.go
rminnich: multierror is not as good an API as I hoped. Even their basic examples fail. I'll stick with it for not but it's a disappointment.

cmds/cpu/cpu9p_unix_test.go
rminnich: The stateful changes reflect the stateful file change. The problem is the way the mask struct is layed out with its many bool fields, as opposed to a bit mask you could easily modify in for loop. You're either going to get a gigantic test full of struct inits, as you progressively enable more and more mask bits; or you're going to end up using reflect to select fields one by one to set them from false to true. I've laid out code for all cases and believe it or not this is the least wordy version. I'd argue that it's also the least clever, as opposed to using reflect. For now, I don't see the need to make this change.

orangecms (approved):

cmds/cpu/cpu9p_unix.go
```diff
@@ -0,0 +1,81 @@
...
+			mtime = time.Unix(int64(attr.MTimeSeconds), int64(attr.MTimeNanoSeconds))
+		}
+		if e := os.Chtimes(l.path, atime, mtime); e != nil {
+			err = multierror.Append(e)
+		}
+	}
+
+	if mask.CTime {
+		// The Linux client sets CTime. I did not even know that was allowed.
+		// Change the create time of a file?
```
brho, line 61: ctime is "change time"

```

       Last status change timestamp (ctime)
              stat.st_ctime; statx.stx_ctime

              This is the file's last status change timestamp.  It is
              changed by writing or by setting inode information (i.e.,
              owner, group, link count, mode, etc.).
```

creation time is birth time (btime)

https://man7.org/linux/man-pages/man7/inode.7.html

cmds/cpu/cpu9p_unix_test.go
```diff
@@ -0,0 +1,199 @@
...
+		ATimeNotSystemTime: false,
+		MTimeNotSystemTime: false,
+	}
+	sa := p9.SetAttr{
+		Permissions:      0777,
+		UID:              ga.UID,
+		GID:              ga.GID,
+		Size:             ga.Size + 4,
+		ATimeSeconds:     ga.ATimeSeconds + 22,
+		ATimeNanoSeconds: ga.ATimeNanoSeconds + 33,
```
brho, line 72: might go over 10^9 ns.  not sure how things will react to that.  could drop a second, and your DeepEqual check will fail later.

PR #90, opened 2022-04-19
proposed table driven test for cpu9p_unix

rminnich:

### DIFF
--- a/cmds/cpu/cpu9p_unix_test.go
+++ b/cmds/cpu/cpu9p_unix_test.go
@@ -197,3 +197,146 @@ func Test9pUnix(t *testing.T) {
 	}
 
 }
+
+func Test9pUnixTable(t *testing.T) {
+	d := t.TempDir()
+	f := filepath.Join(d, "a")
+	if err := ioutil.WriteFile(f, []byte("hi"), 0666); err != nil {
+		t.Fatalf(`ioutil.WriteFile(%q, "hi", 0666): %v != nil`, f, err)
+	}
+	// First test: do nothing
+	m := p9.AttrMask{
+		Mode:        true,
+		NLink:       true,
+		UID:         true,
+		GID:         true,
+		RDev:        true,
+		ATime:       true,
+		MTime:       true,
+		CTime:       true,
+		INo:         true,
+		Size:        true,
+		Blocks:      true,
+		BTime:       true,
+		Gen:         true,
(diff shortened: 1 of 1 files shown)


rminnich:
If you prefer this style I'm fine with it too

cmds/cpu/cpu9p_unix_test.go
```diff
@@ -197,3 +197,146 @@ func Test9pUnix(t *testing.T) {
...
+	}
+
+	c := &cpu9p{
+		path: f,
+	}
+
+	_, _, ga, err := c.GetAttr(m)
+	if err != nil {
+		t.Fatalf("Getattr: %v != nil", err)
+	}
```
orangecms, line 232: Stop here, case is done. Create another case.

cmds/cpu/cpu9p_unix_test.go
```diff
@@ -197,3 +197,146 @@ func Test9pUnix(t *testing.T) {
...
+		"Setattr with size",
+		"Setattr with UID",
+		"Setattr with GID",
+		"Setattr with ATime to local system time",
+		"Setattr with MTime to local system time",
+		"Settattr with ATime to time in TSetattr",
+		"Settattr with MTime to time in TSetattr",
+	}
+	for i, msg := range tests {
+		switch i {
```
orangecms, line 268: This should be a ton easier. Now the test has logic for data; that'd be the point though, to separate the two.

Can we create a function doing something like "give me a modified sam with `Permissions` set to `true`" in Go?

orangecms: Well, and that function should accept any modification.
Can we pass a function that the function executes, doing said modification?

orangecms: In other words: I want the test suite to look just like data.

Data defining what each case means; can we put those switch statements into `tests`?

orangecms: Reading through https://medium.com/@geisonfgfg/functional-go-bc116f4c96a4#:~:text=Golang%20is%20a%20multi%2Dparadigm,easier%20to%20understand%20and%20test. suggests that Go lacks the necessary primitives.

I'm sorry, I was hoping that'd be easier. It would be possible to hack, but that'd only make things worse, I think.

PR #96, opened 2022-04-25
cpud: Always use package code for a session

rminnich:
For the -remote switch, drop all the code that is
not package related.

This is tested for all cases include a kexec of itself.


rminnich:
The next one, which is the big one, is removing session code in favor of the package.

brho (approved):

PR #97, opened 2022-04-26
Cpuinitserverpackage

rminnich:
Please test, I am testing, it seems to work, but let's see. I'll keep pruning and cleaning.


brho:
works for me (chrooting into the target machine with -9p and an fstab)

brho (approved):

PR #98, opened 2022-04-27
Fix the missing -fstab switch on the new package code

rminnich:
We have been doing parallel development and this got lost.

### DIFF
--- a/cmds/cpu/cpu.go
+++ b/cmds/cpu/cpu.go
@@ -439,6 +439,11 @@ func newCPU(host string, args ...string) error {
 	if len(*cpudCmd) > 0 {
 		c.WithCpudCommand(*cpudCmd)
 	}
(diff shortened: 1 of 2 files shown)


cmds/cpu/cpu.go
```diff
@@ -439,6 +439,11 @@ func newCPU(host string, args ...string) error {
 	if len(*cpudCmd) > 0 {
 		c.WithCpudCommand(*cpudCmd)
 	}
+	if len(*fstab) > 0 {
```
rminnich, line 442: honestly, the if is not really needed, but well ... no harm.

brho: can see either way.  you've got the len check elsewhere already

brho (approved):

PR #100, opened 2022-04-29
Continuecleanup

rminnich:
This brings in all the changes in the old cpu that ran ahead of the packages.

Next week it's start writing tests. All kinds of tests.


client/cpu9p_unix.go
```diff
@@ -0,0 +1,82 @@
...
+			mtime = time.Unix(int64(attr.MTimeSeconds), int64(attr.MTimeNanoSeconds))
+		}
+		if e := os.Chtimes(l.path, atime, mtime); e != nil {
+			err = multierror.Append(e)
+		}
+	}
+
+	if mask.CTime {
+		// The Linux client sets CTime. I did not even know that was allowed.
+		// Change the create time of a file?
```
brho: ctime isn't create time.  it's the time the file was changed.  "create time" is btime (birth).

client/cpu9p_unix_test.go
```diff
@@ -0,0 +1,199 @@
...
+		ATimeNotSystemTime: false,
+		MTimeNotSystemTime: false,
+	}
+	sa := p9.SetAttr{
+		Permissions:      0777,
+		UID:              ga.UID,
+		GID:              ga.GID,
+		Size:             ga.Size + 4,
+		ATimeSeconds:     ga.ATimeSeconds + 22,
+		ATimeNanoSeconds: ga.ATimeNanoSeconds + 33,
```
brho: it's possible that these nsec values will be greater than 10^9.  not sure if that will be off-by-one second, or if whoever consumes this will handle it.  so if the test flakes (if it ever checks nsec), that might be why.

brho (approved):

brho (approved):

client/cpu9p_unix_test.go
rminnich: yeah, I decided not to bother checking ns, unless I somehow did so by accident.

client/cpu9p_unix.go
rminnich: fixed.

brho:
thanks for the fixes!

PR #103, opened 2022-05-06
Remove the client 9p timeout

rminnich:
The 9p timeout was originally a way to ensure that the cpud connecting
was the cpud we wanted to connect. It was a very poor security
mechanism.

The use of private fdno mounts and CPUNONCE eliminate the need for a timeout.

In most cases, nowadays, the timeout was causing problems, rather than
showing problems, as it is basically a race condition: the client
has to set up the socket, and wait for a connect; meanwhile, client
must also start up the cpud as fast as it can, before the timeout fires!

It was useful in its time, but the timeout is more of a headache than
anything. Further, it can result in a case where we are running a
remote session, even though the mounts failed, which is a worst
case scenario.

### DIFF
--- a/client/srv.go
+++ b/client/srv.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"log"
 	"net"
-	"time"
 
 	"github.com/hugelgupf/p9/p9"
 	"github.com/u-root/u-root/pkg/ulog"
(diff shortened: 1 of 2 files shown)


brho (commented):
lgtm

brho:
though not sure if the merging happened correctly?  i see there's a merge from main into your branch, instead of a rebase.  not a big deal.

PR #109, opened 2022-06-29
Make With* idiomatic Go

rminnich:
At some point I got into Rust habits (in part because of a Go code review
on a totally different project, as it happens) and was using a more Rust style
for options:
c := client.Command(...).WithThis(a).WithThat(b)

That's not really how it should be done, and I had forgotten, but a look
at u-root (see the find package in u-root) made me realize the mistake.

Further, unlike Rust, failing out of those With* things gets nasty,
because Go doesn't have anything like the Rust Result and things like
the Rust ? operator. In one case, I've seen people scarfing the error
value away in the struct, and checking the error and exiting With functions
instantly if the error is set, and checking the error in the struct at the end,
and using that as an error value, and oh please make it stop.

This is not to say one way is unambigously better.
Rust has greater complexity to go with its extra capability.
Go has greater simplicity to go with its slightly greater inconvenience.
Go makes writing and using new Set functions easy, in or out of the package.

Anyway:
Add the Set type to the client package:
type Set func(*Cmd) error

Change the With functions and use as follows:
c := client.Command(host, args...)

Add a new function, SetOptionsin client:
func (c*Cmd) SetOptions(...Set) error

to be used as follows:
if err := c.SetOptions(client.WithThis, client.WithThat, ...); err != nil

In actual use, this turns out to be an improvement.

This is all in preparation for enabling use of the vsock network type.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -73,6 +73,16 @@ type Cmd struct {
 	closers []func() error
 }
 
+// SetOptions sets various options into the Command.
+func (c *Cmd) SetOptions(opts ...Set) error {
+	for _, o := range opts {
+		if err := o(c); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // Command implements exec.Command. The required parameter is a host.
 // The args arg args to $SHELL. If there are no args, then starting $SHELL
 // is assumed.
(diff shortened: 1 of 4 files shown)


brho (approved):
as they say, all things can be solved with another layer of (func/closure) indirection!

PR #114, opened 2022-07-25
Add vsock support to cpud and cpu client package

rminnich:
vsock is a simpler networking path for virtual machine guests.

It needs no emulated hardware or IP stacks. It is a lot like
networking over a serial line. There are fewer moving parts
and hence less to go wrong.

One thing that can go wrong is your kernel is broken, e.g. ubuntu
5.8, and in that case you will get connection timeouts. To fix
a problem like this, update ubuntu or use something else. Ubuntu
22 is known to work; ubuntu 20 is known to not work.

Sadly, vsock is not in the net package, so the code is more complex
than it ought to be.

To use vsock on cpud:
cpud -net vsock [... args]

cpud will listen on any CID, so can be reached even from the VM itself.

To use it on the client:
./cpu -net vsock [other switches] vsock-client-id [other cpu args]

e.g., if one starts qemu with the following vsock switches:
-device vhost-vsock-pci,guest-cid=42

The vsock-client-id above would be 42.

For example, when qemu is started as above, then the cpud command is
cpud -net vsock -init

and the cpu command can be (assuming you are not using a default
key):
./cpu -key ~/.ssh/cpu_rsa -net vsock 42 date

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-multierror"
+	"github.com/mdlayher/vsock"
 	"github.com/u-root/u-root/pkg/termios"
 	"golang.org/x/crypto/ssh"
 )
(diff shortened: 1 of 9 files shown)


client/client.go
```diff
@@ -259,8 +260,30 @@ func (c *Cmd) Dial() error {
...
+	// Sadly, no vsock in net package.
+	var (
+		cl   *ssh.Client
+		err  error
+		addr string
+	)
+
+	switch c.network {
+	case "vsock":
+		id, port, err := vsockIdPort(c.HostName, c.Port)
```
brho, line 272: do you need to check err on this line?

client/client.go
```diff
@@ -259,8 +260,30 @@ func (c *Cmd) Dial() error {
...
+		conn, err := vsock.Dial(id, port, nil)
+		if err != nil {
+			return err
+		}
+		addr = fmt.Sprintf("%#x:%d", id, port)
+		c, chans, reqs, err := ssh.NewClientConn(conn, addr, &c.config)
```
brho: i'm not familiar with the ssh package, but would it be worth having all cases (vsock, inet) operate on chans and addrs, and then call NewClientConn and NewClient outside the switch statement?

client/client.go
rminnich: yeah, thanks, and I added a test for vsockIdPort

client/client.go
rminnich: yeah, that's a great idea, see what you think of this now.

brho (approved):

PR #115, opened 2022-07-25
Add an "inetd" mode for cpud

rminnich:
People are looking at running cpud manually, after the system is fully booted.

cpud will now have a default behavior when it is invoked with no mode: it will
run as "inetd." It will not try to set up cgroups, loopback interfaces, or a
process reaper: that is for init to do.

This will make it easier to run cpud on systems which are past runlevel 1.

### DIFF
--- a/cmds/cpud/serve.go
+++ b/cmds/cpud/serve.go
@@ -34,10 +34,7 @@ func hang() {
 	log.Printf("done hang")
 }
 
-func serve() error {
-	if err := unix.Mount("cpu", "/tmp", "tmpfs", 0, ""); err != nil {
-		log.Printf("CPUD:Warning: tmpfs mount on /tmp (%v) failed. There will be no 9p mount", err)
-	}
+func commonsetup() error {
 	if *debug {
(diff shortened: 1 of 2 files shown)


brho (approved):

cmds/cpud/serve.go
```diff
@@ -29,7 +29,7 @@ func hang() {
 	log.Printf("done hang")
 }
 
-func serve() error {
+func commonsetup() error {
 	if err := unix.Mount("cpu", "/tmp", "tmpfs", 0, ""); err != nil {
```
brho: oh, this might not be good for systems that are up and running.

brho (changes requested):

cmds/cpud/serve.go
brho: i think this will work if you move this Mount to initsetup()

brho:
did you want to do the tmpfs mount in the -init case still?

brho (approved):

PR #116, opened 2022-07-25
Change all uses of 127.0.0.1 with "localhost"

brho:
For systems with no IPv4 stack.

In particular, the client.go connection tells cpud to connect to
127.0.0.1.  But if there is no 127.0.0.1, you'll get an error like:

	SSH error Dial: cpu client listen for forwarded 9p port ssh:
	tcpip-forward request denied by peer

Signed-off-by: Barret Rhoden <brho@google.com>

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -276,7 +276,7 @@ func (c *Cmd) Dial() error {
 	// Note: cl.Listen returns a TCP listener with network "tcp"
 	// or variants. This lets us use a listen deadline.
 	if c.Ninep {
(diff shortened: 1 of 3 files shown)


brho:
not 100% this is right - i think if i tunnel cpu's initial connection over an ssh -L port forward, the dialback still doesn't work if cpud is in a v6-only environment.

brho:
yeah, i think this doesn't work if you are tunnelling your cpu -> cpud connection over ssh or something.  but if you manually do the fallback from ipv4 to ipv6 in the listen code, it'll work.  i'll open another PR for that.

PR #118, opened 2022-07-26
more cpud stuff

brho:
- more ipv6 support
- /proc/self/exe on linux to find cpud.

tested with weird ssh tunnels with cpud not having IPv4, and cpud not in the $PATH.

### DIFF
--- a/client/cpu_linux.go
+++ b/client/cpu_linux.go
@@ -11,6 +11,8 @@ import (
 	"github.com/hugelgupf/p9/p9"
 )
 
+const osDefaultCpudCmd = "/proc/self/exe -remote"
+
 func osflags(fi os.FileInfo, mode p9.OpenFlags) int {
 	flags := int(mode)
 	if fi.IsDir() {
--- a/server/server_other.go
+++ b/server/server_other.go
@@ -27,7 +27,7 @@ func (s *Session) Namespace() (error, error) {
 	v("CPUD:namespace is %q", s.binds)
 
 	// Connect to the socket, return the nonce.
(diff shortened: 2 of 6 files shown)


client/cpu_linux.go
```diff
@@ -18,3 +18,7 @@ func osflags(fi os.FileInfo, mode p9.OpenFlags) int {
 	}
 	return flags
 }
+
+func osDefaultCpudCmd() string {
```
rminnich: fwiw: these could also be 
const osDefaultCpudCmd = "/proc/self/exe -remote"
for example, i.e. no need to be a function, wdyt?

server/server_other.go
```diff
@@ -27,7 +27,7 @@ func (s *Session) Namespace() (error, error) {
 	v("CPUD:namespace is %q", s.binds)
 
 	// Connect to the socket, return the nonce.
-	a := net.JoinHostPort("127.0.0.1", s.port9p)
+	a := net.JoinHostPort("localhost", s.port9p)
```
rminnich, line 30: the reason I used the raw ip was because /etc/hosts is not guaranteed to be there, but, are you doing this for ip6 / ip4 reasons? If so, I think I can drop my paranoia :)

client/cpu_linux.go
brho: done, thanks

rminnich (approved):

server/server_other.go
brho: yeah, ipv4 isn't always there.  and i think "localhost" is handled by the go runtime magically, without involving /etc/hosts.

brho: i think

PR #120, opened 2022-07-26
Maybe deal with PWD is not a directory in the remote namespace

brho:
PWD on the cpu client side might not exist in the namespace created by
cpud -remote.  For instance, if you run cpu from /tmp/foo, and that is
not exported in the 9p namespace= parameter, the command will fail with
something like:

	CPUD(as remote):chdir /tmp/foo: no such file or directory

Signed-off-by: Barret Rhoden <brho@google.com>

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -200,6 +200,11 @@ func (s *Session) Run() error {
 	v("CPUD:runRemote: command is %q", s.args)
 	c := exec.Command(s.cmd, s.args...)
 	c.Stdin, c.Stdout, c.Stderr, c.Dir = s.Stdin, s.Stdout, s.Stderr, os.Getenv("PWD")
+	dirInfo, err := os.Stat(c.Dir)
+	if err != nil || !dirInfo.IsDir() {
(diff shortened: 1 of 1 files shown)


rminnich (approved):
nice checking the IsDir thing :-)

rminnich:
let's think about this a bit more?

brho:
yeah, i think maybe not for this one.  too many subtleties.  i'll change it to be a more verbose error.

brho:
changed it to print a more verbose notice.  though not sure if this is excessive or not.

still wonder if we should put PWD into the *default* namespace argument.

rminnich:
thinking about it ... I think this cure may be worse than the disease. maybe the real fix is just a more useful error message: have the session stat ($PWD) and point out that it's not in the namespace?

brho:
yeah, the loud error doesn't break anything, and should at least help diagnose the issue.  i updated the patch to just print and error out.

rminnich (approved):

rminnich (approved):

PR #121, opened 2022-07-28
Revert "Use /proc/self/exe on Linux for cpud"

rminnich:
Using the /proc/self/exe symlink is a nice idea, but on u-root systems
it resolves to /bbin/bb, which is absolutely not what we want.

This reverts commit c0075b1ce7e8db5669f70b08731e986a397b7c12.


brho:
+2

PR #124, opened 2022-08-01
cmds/cpu: remove termios calls

rminnich:
Now that the u-root termios pkg is fixed as of b0da363052a2724a46a72ec11d323e6f646529a9,
we no longer need the calls in cpu.go to get and reset the terminal.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -58,7 +58,7 @@ type Cmd struct {
 	Stderr         io.Reader
 	Row            int
 	Col            int
-	interactive    bool // Set if there are no arguments.
+	hasTTY         bool // Set if we have a TTY
 	// NameSpace is a string as defined in the cpu documentation.
 	NameSpace string
 	// FSTab is an fstab(5)-format string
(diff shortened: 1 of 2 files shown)


brho (approved):

client/client.go
```diff
@@ -100,7 +100,10 @@ func Command(host string, args ...string) *Cmd {
 
 	col, row := 80, 40
 	if w, err := termios.GetWinSize(0); err != nil {
-		V("Can not get winsize: %v; assuming %dx%d", err, col, row)
+		// This means it is not a tty, and hence for all intents and
+		// purposes, not interactive.
+		interactive = false
+		log.Printf("Can not get winsize: %v; assuming %dx%d and non-interactive", err, col, row)
 	} else {
 		col, row = int(w.Col), int(w.Row)
```
Lencerf, line 105: I think we need `interactive = true` here.

Lencerf (approved):

PR #125, opened 2022-08-02
Add the dut command from linuxboot

rminnich:
It is much easier to bring dut into cpu than have 2 repos.
Further, dut is not a combined cpu and cpud; it is useful
to have this illustrative use of the cpu package.


cmds/dut/doc.go
```diff
@@ -0,0 +1,66 @@
...
+// they follow that argument.
+// E.g., when uinit is run on the host and we want it to enable cpu daemons
+// on the DUT, we run it as follows:
+// dut cpu -key ...
+// the -key switch is only valid following the cpu mode argument.
+//
+// modes
+// dut currently supports 3 modes.
+//
+// The first, default, mode, is "device". In device mode, dut makes an http connection
```
brho, line 25: is this "In device mode, DUT makes"  vs dut makes?  or dut on DUT makes?

also, should the DUT start the RPC server before calling back to the host?  if you do it as written, there's a window of time where the host dut could call back to the DUT, but the service isn't ready yet.  if you time it poorly, you'll get stuck in tcp limbo or something.

cmds/dut/doc.go
```diff
@@ -0,0 +1,66 @@
...
+// to a dut running on a host, then starts an  HTTP RPC server.
+//
+// The second mode is "tester". In this mode, dut calls the Welcome service, followed
+// by the Reboot service. Tester can be useful, run by a shell script in a for loop, for
+// ensure reboot is reliable.
+//
+// The third mode is "cpu". dut will direct the DUT to start a cpu service, and block until
+// it exits. Flags for this service:
+// pubkey: name of the public key file
```
brho, line 34: is there a nice Go way to use the same parameters as the cpud command?  e.g. right now cpud has pubkey == -pk.  and of course -net for vsock.

not sure if we can have a package struct that is the expected configuration, etc, so that we don't have two copies of all the args and parsing that cpud expects.  as an added bonus, we wouldn't have different names for the same arg either.

cmds/dut/doc.go
```diff
@@ -0,0 +1,66 @@
...
+// dut currently supports 3 modes.
+//
+// The first, default, mode, is "device". In device mode, dut makes an http connection
+// to a dut running on a host, then starts an  HTTP RPC server.
+//
+// The second mode is "tester". In this mode, dut calls the Welcome service, followed
+// by the Reboot service. Tester can be useful, run by a shell script in a for loop, for
+// ensure reboot is reliable.
+//
+// The third mode is "cpu". dut will direct the DUT to start a cpu service, and block until
```
brho, line 32: regarding "block until" - i might want the host dut command to return once the cpud service is running.  i.e. "the CPU RPC is complete when i know the server is running", and not "when the server has exited".  

this will help me with vmapp, where i can block until i know cpud is fully listening.

dut/doc.go
```diff
@@ -0,0 +1,66 @@
+// dut manages Devices Under Test (a.k.a. DUT) from a host.
```
brho: this looks like a copy of the cmds/dut/doc.go.  from glancing at the Changed Files, it looks like dut/ and cmds/dut/ are the same.  not sure which one you want to keep.

brho (commented):
i haven't looked at the actual code yet - comments just on the Doc and method of operation stuff.

cmds/dut/doc.go
rminnich: can we have common flags in the client/ package? probably.

PR #129, opened 2022-08-29
client/client.go: Fix the vsock listen bug AGAIN

rminnich:
Need to predeclare two variables.

Signed-off-by: Ronald G. Minnich <rminnich@google.com>

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -253,6 +253,20 @@ func WithPort(port string) Set {
 
 }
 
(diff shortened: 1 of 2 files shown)


brho (approved):
doh!

would it help to pull that switch statement out of Dial() and put it in a helper function, and avoid using the err declared at the top of Dial()?  similar to your previous fix where you made a function that returned a listener or something. 

if not, i'm fine with it as is.

PR #130, opened 2022-09-02
docker cleanup and add multiarch

ericvh:
This moves the build into the Dockerfile and cleans up stuff that
isn't necessary in the new setup.
This also allows for clean multi-arch builds and how to do it is
added to the README.md


ericvh:
I'm not skilled in CircleCI, but it seems like it should be possible to have CircleCI autobuild an official multi-arch Docker container.  I have instructions for doing this on gitlab, not sure if they'll work in default CircleCI environment though.

rminnich (approved):

PR #131, opened 2022-09-06
Add dnssd support to cpu

ericvh:
First draft of mdns support to cpu and cpud

This is still a WIP, but is functional - it just selects the first mdns response which I want to expand to allow for selection criteria and prioritization criteria to allow the user to say they want to CPU into the lowest-load arm based cpu-server.

I probably need some help figuring out the right incantations to incorporate dependencies into the .circleci build - my build is based on my Dockerfile and .gitlab-ci.yml


cmds/cpud/serve.go
```diff
@@ -91,14 +95,20 @@ func listen(network, port string) (net.Listener, error) {
 			return nil, err
 		}
 		ln, err = vsock.ListenContextID(any, uint32(p), nil)
-
+		if err != nil {
```
rminnich: This return is not needed. 

There seem to be several different things going on in this PR, can you split out into multiple PRs. It makes bisect way easier.

cmds/cpud/serve.go
```diff
@@ -115,6 +125,31 @@ func serve() error {
 	if err != nil {
 		return err
 	}
+	// TODO: This is overly restrictive
```
rminnich: This should be done in the server package.

cmds/cpu/cpu.go
```diff
@@ -170,12 +175,32 @@ func newCPU(host string, args ...string) error {
 
 func main() {
 	flags()
+	host := ""
 	args := flag.Args()
-	if len(args) == 0 {
-		usage()
+	a := ""
+
+	if *mdnsEnabled {
```
rminnich: This probably needs to be in the client package.

rminnich (commented):
There's def one thing you don't need here. Let's start with that.

The way this is done will make it hard to write tests, so let's figure that out too. We don't have enough coverage as it is, so we don't want to make  it worse.

cmds/cpud/serve.go
```diff
@@ -91,14 +95,20 @@ func listen(network, port string) (net.Listener, error) {
...
+		}
 	case "unix", "unixgram", "unixpacket":
 		// net.JoinHostPort really ought to work for UDS, but it's very naive.
 		// It does not take the network type as a parameter.
 		ln, err = net.Listen(network, port)
 
 	default:
 		ln, err = net.Listen(network, net.JoinHostPort("", port))
+		if err != nil {
+			return nil, err
```
rminnich: This test is not needed.

cmds/cpud/serve.go
```diff
@@ -91,14 +95,20 @@ func listen(network, port string) (net.Listener, error) {
...
+		}
+
 	}
 	return ln, err
```
rminnich: Evreryone returns here.

cmds/cpud/serve.go
ericvh: removed

cmds/cpud/serve.go
```diff
@@ -115,6 +118,31 @@ func serve() error {
 	if err != nil {
 		return err
 	}
+	// TODO: Move to server package
+	if *mdnsEnabled {
+		v("starting mdns server")
+		// Setup our service export
+		host, _ := os.Hostname()
+		var p uint64
+		p, err = strconv.ParseUint(*port, 10, 32)
```
rminnich: Are ports 32 bits? Do we need to require base 10? People might want to say port 0xcafe maybe?

cmds/cpud/serve.go
rminnich: Could we indicate mdns in the network address somehow and let client code figure it out? 
What if as part of hostname parsing we could just magically now on the client side what to do? What if on server side we could, similarly, know that cpud needs to be an mdns server? There seem to be some kind of standards out there for it. I really want to get this in, it would be super handy, just trying to work out how. 

Can we just mount an mdns server on /net/cs -- oh wait ...

ericvh: Yeah, I did think that we could do magic hostname specification -- of course that always comes with its own caveats.  Server side autodetecting if it has to act as server seems a bit more dodgey - in general things seem to work on metal, but docker won't let multiple people control the port.  There seem to be multiple things out there to make docker play nice with mdns, but haven't had time to really try any of them in anger.  In any case, probably will need to make that different variations of options (host your own server, play nice with dbus, use our own mdns port, do something else)

Sorry - was a busy week this last week, so I didn't get around to doing much here - was planning on spending some time this week implementing the code reorg into the modules like you suggested and fleshing out the meta-data based masking and prioritization.

rminnich: very cool, I was surprised to see this closed but if you're going to keep at it, we REALLY need it. It's going to make life a lot better for our use cases.

ericvh:
@rminnich - pushed movement of mdns into client module -- this relies on you being able to set network to mdns and using hostname to specify selection/prioritization criteria.  This pass muster or do you want me to try and parse URI to specify mdns -- and then what kind of syntax:
mdns://arch=arm64,mem>16G:cpucount>16 
or
mdns://_ncpu._tcp(arch=arm64,mem>16G:cpucount>16)
or
mdns!_ncpu._tcp!_arch=arm64,mem>16G:cpucount>16
(or something else)

Of course, this may be the next level of abstraction - maybe just mdns to start with as hostname?

cmds/cpud/serve.go
ericvh: Agreed.  will stop specifying the port conversion.

cmds/cpu/cpu.go
ericvh: @rminnich - pushed movement of mdns into client module -- this relies on you being able to set network to mdns and using hostname to specify selection/prioritization criteria. This pass muster or do you want me to try and parse URI to specify mdns -- and then what kind of syntax:
mdns://arch=arm64,mem>16G:cpucount>16
or
mdns://_ncpu._tcp(arch=arm64,mem>16G:cpucount>16)
or
mdns!_ncpu._tcp!_arch=arm64,mem>16G:cpucount>16
(or something else)

Of course, this may be the next level of abstraction - maybe just mdns to start with as hostname?

cmds/cpud/serve.go
ericvh: server portion migrated to server module, but it wasn't as clean as I would like.  Suggestions welcome.

server/server.go
```diff
@@ -105,6 +108,28 @@ func handler(s ssh.Session) {
 	verbose("handler exits")
 }
 
+func Advertise(hostname, port, txt string) (*mdns.Server, error) {
+	v("starting mdns server")
+	// Setup our service export
+	host, _ := os.Hostname()
```
abrender: Is it safe to ignore the error here? I think you might want to handle it?

server/server.go
```diff
@@ -105,6 +108,28 @@ func handler(s ssh.Session) {
...
+		return nil, err
+	}
+	service, err := mdns.NewMDNSService(host, "_ncpu._tcp", "", "", int(p), nil, strings.Split(txt, ","))
+	if err == nil {
+		// Create the mDNS server, defer shutdown
+		// we have to lookup the right interface, this will have to be passed in from command line
+		server, err := mdns.NewServer(&mdns.Config{Zone: service})
+		return server, err
+	} else {
+		log.Printf(`mdns new service failure: %v`, err)
```
abrender: Do you want to return an error here instead of logging and returning err = nil?

I'm trying to understand how the code is structured - is the goal here to still return a nil error if mdns.NewMDNSService() fails?    If not then maybe you can simplify and restructure the code as:
```
service, err := ....
if err != nil {
  // Handle the error
}
return mdns.NewServer(&mdns.Config{Zone: service})
```
?

server/server.go
abrender: (Specifically referring to the error returned by `os.Hostname()`)

server/server.go
ericvh: the original goal was to try and keep mdns as optional as possible so if mdns resolution fails, just fall through and hope the user specified the hostname.  I figured the end result would be a host not found if otherwise -- but I'm happy to take action to make this more explicit or specific (no available resource found versus host not found)

server/server.go
ericvh: It's a fair comment.  In some ways I was trying to follow lighter weight error handling focus elsewhere in the code -- but this sort of error likely to cause issues so maybe better explicitly report and then bail out.

server/server.go
abrender: Thanks for letting me know the goal, that is helpful.   In this case I would think that it's still better to explicitly return the error here from `Advertise` in server.go (because we are failing to advertise...), but ignore the error in the *caller* of the function, which appears to be cpud/serve.go in the `if *mdnsEnabled{` section of code - and it appears you're already doing that.

So properly returning the errors here won't change the existing optionality that is already accomplished by the caller in cpud/serve.go ignore the returned error.

Please let me know if I've misunderstood? Thanks!

cmds/cpud/serve.go
```diff
@@ -115,6 +114,12 @@ func serve() error {
 	if err != nil {
 		return err
 	}
+
+	if *mdnsEnabled {
+		host, _ := os.Hostname()
+		mdns_server, _ := server.Advertise(host, *port, *mdnsTxt)
```
abrender: I understand that the intent here is to make this optional so the error is ignored. But future readers of the code might not understand this, so can you please add a comment about why the error is ignored? Thanks!

cmds/cpud/serve.go
```diff
@@ -115,6 +114,12 @@ func serve() error {
...
+		defer mdns_server.Shutdown()
```
abrender: calling `mdns_server.Shutdown()` probably won't work if `server.Advertise` returned an error, right? So maybe you can structure the code as

```
// Comment here explaining why we will ignore the `err != nil` case.
if mdns_server, err := server.Advertise(...); err == nil {
  defer mdns_server.Shutdown()
}
```

ericvh: actually, reflecting on this a bit more, if the user specifies mdns advertisment on the command line and it goes sideways, I think we just treat it as failure.  As long as mdns advertisement isn't default, this should be fine.  I'll restructure code appropriately.

server/server.go
ericvh: I've added back in error handling and made failure less silent if you selected the mdns advertisement option.

client/client.go
```diff
@@ -267,6 +269,80 @@ func vsockDial(host, port string) (net.Conn, string, error) {
...
+
+// check that mdns response has all required attributes
+func mdns_required(src []string, req map[string][]string) bool {
+	txt := make(map[string]string)
+	for element := range src {
+		s := strings.SplitN(src[element], "=", 2)
+		txt[s[0]] = s[1]
+	}
+	for k, _ := range req {
+		if !contains(req[k], txt[k]) {
```
abrender: I think you can use https://pkg.go.dev/golang.org/x/exp/slices#Contains here.  It is my understanding that this is quite possibly going to be merged into the standard library in the future.

Note that both the above package and the current implementation are O(N) which is expensive... if you use a map then it's O(1).  I don't know the size of these slices to know if that's important enough though... so I think the performance implications of the slower implementation should be ignored in favor of the simplicity of using https://pkg.go.dev/golang.org/x/exp/slices#Contains

client/client.go
```diff
@@ -267,6 +269,80 @@ func vsockDial(host, port string) (net.Conn, string, error) {
...
+	var (
+		hostname string
+		port     string
+		network  string
+		err      error
+	)
+
+	// determine if we want to do mdns
+	u, err := url.Parse(uri)
+	if (err == nil) && (u.Scheme == "dnssd") {
```
abrender: you can save an entire layer of indent for the following ~30 lines of code if you structure the code as:
```
if err != nil {
  // return code
}
if u.Scheme != "dnssd") {
  return "", "", "", nil // Nothing to do here...
}
// rest of code.
```

client/client.go
```diff
@@ -267,6 +269,80 @@ func vsockDial(host, port string) (net.Conn, string, error) {
...
+		}
+		if u.Path != "" {
+			mdns_type = u.Path
+		}
+
+		mdns_txt := u.Query()
+
+		// Make a channel for results and start listening
+		entriesCh := make(chan *mdns.ServiceEntry, 64)
+		// Setup sensible defaults"
```
abrender: Is the quotation mark at the end of this comment a typo?

client/client.go
```diff
@@ -267,6 +269,80 @@ func vsockDial(host, port string) (net.Conn, string, error) {
...
+			mdns_domain = u.Host
+		}
+		if u.Path != "" {
+			mdns_type = u.Path
+		}
+
+		mdns_txt := u.Query()
+
+		// Make a channel for results and start listening
+		entriesCh := make(chan *mdns.ServiceEntry, 64)
```
abrender: Where does the number 64 come from?

client/client.go
```diff
@@ -267,6 +269,80 @@ func vsockDial(host, port string) (net.Conn, string, error) {
...
+		// Setup sensible defaults"
+		// Start the lookup
+		err = mdns.Lookup(mdns_type+"."+mdns_domain, entriesCh)
+		if err != nil {
```
abrender: I'm confused by this, should it be `if err == nil {` ? The following code ranges through the entriesCh channel, but if `err != nil` then I'd not expect this to work?

client/client.go
```diff
@@ -267,6 +269,80 @@ func vsockDial(host, port string) (net.Conn, string, error) {
...
+			for entry := range entriesCh {
+				V("Got new mdns entry: %v\n", entry)
+				if mdns_required(entry.InfoFields, mdns_txt) {
+					hostname = entry.Addr.String()
+					port = strconv.Itoa(entry.Port)
+					network = "tcp"
```
abrender: is `network` the right name for this variable? Should it be `protocol`?

cmds/cpud/serve.go
```diff
@@ -115,6 +116,15 @@ func serve() error {
 	if err != nil {
 		return err
 	}
+
+	if *mdnsEnabled {
+		mdns_server, err := server.Advertise(*port, *mdnsTxt)
+		if err != nil {
+			return fmt.Errorf("Could not advertise with mdns: %w", err)
+		} else {
```
abrender: I don't think this `else` is needed because the previous `if` statement `return`s
```
if err != nil {
  return fmt.Errorf(....)
}
defer mdns_server.Shutdown()
```

server/server.go
```diff
@@ -105,6 +108,27 @@ func handler(s ssh.Session) {
...
+	p, err := strconv.ParseUint(port, 0, 0)
+	if err != nil {
+		return nil, fmt.Errorf("Could not parse port: %s, %w", port, err)
+	}
+	service, err := mdns.NewMDNSService(host, "_ncpu._tcp", "", "", int(p), nil, strings.Split(txt, ","))
+	if err != nil {
+		return nil, fmt.Errorf("Could not instantiate mDNS server on port %d: %w", p, err)
+	}
+
+	server, err := mdns.NewServer(&mdns.Config{Zone: service})
```
abrender: You can do `return mdns.NewServer(...)` here instead of `server, err := ...; return server, err`

client/client.go
ericvh: Thanks for the pointer.  This shouldn't be in a performance critical piece of code, so it shouldn't be a huge issue.

client/client.go
ericvh: yeah, I waffled between the two, happy to side with not having the indent.

client/client.go
ericvh: well spotted

client/client.go
ericvh: its arbitrary, nearly as arbitrary as the 4 that was there before -- felt like a deeper channel would be less likely to miss options, happy to have suggestions on best practices here.

client/client.go
ericvh: well spotted.  I committed and pushed at the end of the day once things compiled so I could test at home (to be fair, the Pull request is still marked as WIP) -- thanks for early debugging this ;)

client/client.go
ericvh: I was following the paradigm of the client options which callas it network -- I do realize that its unlikely we'd actually get a unix or a vsock response back from mdns...but those are more network than protocol.  It starts to get fuzzy.

cmds/cpud/serve.go
ericvh: true, will simplify.

server/server.go
ericvh: will simplify.

client/client.go
rminnich: it's a go thing ... they use net and network when they kind of mean protocol. This is due to Plan 9 root IIUC.

ericvh:
There are some unfortunate problems due to the lack of functionality (one might even say bugs) in the hashicorp/mdns package. I'm currently investigating swapping it out and will also look at incorporating the experimental slice compare library.

client/client.go
ericvh: So, I took your advice and everything works except for CI because we are using an older golang image.  @rminnich - can we upgrade the image we use for circleCI or should I revert back to my custom contains function?

ericvh: I need at least go 1.18 but the current image tops out at 1.17

ericvh: Had to update the underlying image for CI, but was able to follow your suggestions.

ericvh:
Okay, I'm fairly happy with this as MVP -- let me know if you want me to do a rebase/squash to make it easier to review.

rminnich (approved):

PR #133, opened 2022-09-09
Add -tmpMnt to change namespace mount point

Lencerf:
Some applications may need to access the original contents in `/tmp` directory of the local host. Adding `-tmpMnt` to allow changing the private namespace mount point to avoid conflict.

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path/filepath"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/u-root/cpu/mount"
@@ -43,6 +44,7 @@ type Session struct {
 	port9p string
 	cmd    string
 	args   []string
+	tmpMnt string
 }
 
 var (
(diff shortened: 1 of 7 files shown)


session/session.go
```diff
@@ -102,13 +103,13 @@ func (s *Session) TmpMounts() error {
 	// It's true we are making this directory while still root.
 	// This ought to be safe as it is a private namespace mount.
 	// (or we are started with a clean namespace in Plan 9).
-	for _, n := range []string{"/tmp/cpu", "/tmp/local", "/tmp/merge", "/tmp/root", "/home"} {
+	for _, n := range []string{s.tmpMnt + "/cpu", s.tmpMnt + "/local", s.tmpMnt + "/merge", s.tmpMnt + "/root", "/home"} {
```
rminnich: This needs to use filepath.Join, not +

session/session_linux.go
```diff
@@ -96,8 +96,8 @@ func (s *Session) Namespace() (error, error) {
 	if len(s.mopts) > 0 {
 		opts += "," + s.mopts
 	}
-	v("CPUD: mount 127.0.0.1 on /tmp/cpu 9p %#x %s", flags, opts)
-	if err := unix.Mount("localhost", "/tmp/cpu", "9p", flags, opts); err != nil {
+	v("CPUD: mount 127.0.0.1 on %s/cpu 9p %#x %s", s.tmpMnt, flags, opts)
```
rminnich: use filepath.Join(s.tmpMnt, "cpu")

session/session_linux.go
```diff
@@ -96,8 +96,8 @@ func (s *Session) Namespace() (error, error) {
...
+	if err := unix.Mount("localhost", s.tmpMnt+"/cpu", "9p", flags, opts); err != nil {
```
rminnich: pretty much everywhere you've to any kind of assembling of file names, use filepath.Join

There's a very small chance cpud could run on non-linux systems some day.

session/session.go
Lencerf: Done.

session/session_linux.go
Lencerf: Done

session/session_linux.go
Lencerf: Done.

Lencerf:
@rminnich In cmds/cpud/serve.go, since `initsetup()` is called in the `runAsInit` mode, not in the `-remote` mode, do we need the following tmpfs mount?
https://github.com/u-root/cpu/blob/b8d979804e637869b041e1599ab7b03ffdd31557/cmds/cpud/serve.go#L48-L51
 
If it is running as init, we might need that mount if root is on, e.g., readonly media.

rminnich (approved):

rminnich (approved):

PR #134, opened 2022-09-14
Remove use of p9.unimplfs

rminnich:
It has proven to be very confusing to use unimplfs, since it shortstops T requests silently. This was, when cpu is run with -d, users will see the messages.

### DIFF
--- a/client/cpu9p.go
+++ b/client/cpu9p.go
@@ -18,6 +18,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"syscall"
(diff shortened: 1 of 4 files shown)


client/cpu9p.go
```diff
@@ -282,9 +283,19 @@ func (*cpu9p) Rename(directory p9.File, name string) error {
 }
 
 // RenameAt implements p9.File.RenameAt.
-func (*cpu9p) RenameAt(oldname string, newdir p9.File, newname string) error {
-	verbose("RenameAt: not implemented")
-	return syscall.ENOSYS
+// There is no guarantee that there is not a zipslip issue.
+func (l *cpu9p) RenameAt(oldName string, newDir p9.File, newName string) error {
+	oldPath := path.Join(l.path, oldName)
```
brho, line 288: somewhat interesting that 9p does renameAt on the directory, taking the file name.  versus operating on the target itself.  when i did this in the kernel i did something like
  struct chan *renamee = namec(from_path, Aremove, 0, 0, NULL);
where renamee was the actual file to move, not the parent. 

though there isn't always a 1:1 relationship between struct dev function pointers and 9p calls.

brho (approved):

client/cpu9p.go
rminnich: it's all about the fid. You walk to the old and new dir, then pass them then old and new name. There's a weird consistency to it.

PR #139, opened 2022-09-24
dnssd: add comparison operators to requirements

ericvh:
This allows you to do more than just equivilience checks for metadata requirements.

e.g. dnssd:?mem=>1024   or dnssd?cores=>4 or dnssd?cores=!1

Values still need to be key/value style (still need an equal.


ericvh:
Picked up in a different pull request which was dependent on this one.

PR #146, opened 2022-09-30
client: only pass CPU_FSTAB to cpud

rminnich:
cpu has been able to pass an fstab in CPU_FSTAB for a while.

This led to having two ways to specify a namespace to cpud: CPU_NAMESPACE and CPU_FSTAB.

We can simplify this, and simplify cpud, by having the client parse the namespace and add it to CPU_FSTAB.

Further, if the client parses the contents of -namespace for errors, feedback for bad values is much more
immediate.

This change:
o implements parsing the namespace to CPU_FSTAB
o concatenating it at the end of a user-specified
  fstab file (if any)
o NOT passing CPU_NAMESPACE to cpud

If this works, the next step is to remove CPU_NAMESPACE support from cpud, thereby simplifying it.

The namespace switch will be preserved for the client, because it is very convenient.


client/client.go
```diff
@@ -567,14 +567,23 @@ func DsLookup(query dsQuery) (string, string, error) {
 // Dial implements ssh.Dial for cpu.
 // Additionaly, if Cmd.Root is not "", it
 // starts up a server for 9p requests.
+// Note that any bind parsing is deferred until this point,
```
brho, line 291: i guess Dial() is the first thing called after the cmd args parsing?  wonder if there's a more explicit "finished with args" function we can call, in case later on we do something between the arg parsing and Dial.

client/fns.go
```diff
@@ -252,3 +252,69 @@ func (c *Cmd) Outputs() ([]bytes.Buffer, error) {
 	}
 	return r[:], nil
 }
+
+// parseBinds parses a CPU_NAMESPACE-style string to a
+// an fstab format string. It is not intended to be called
+// outside this package, because it is sensitive to values
+// in the Cmd struct (e.g. c.TmpMnt) and users might set things
```
brho, line 259: is this bit about c.TmpMnt still true?  since you're passing in TmpMnt as a string, and not as 'c' directly.  regardless, keeping it in this package makes sense to me.

client/fns.go
```diff
@@ -252,3 +252,69 @@ func (c *Cmd) Outputs() ([]bytes.Buffer, error) {
...
+		if len(local) == 0 {
+			return fstab, fmt.Errorf("bind: element %d(%q): local is 0 length:%w", i, bind, strconv.ErrSyntax)
+		}
+		if len(remote) == 0 {
+			return fstab, fmt.Errorf("bind: element %d(%q): remote is 0 length:%w", i, bind, strconv.ErrSyntax)
+		}
+
+		// The convention is that the remote side is relative to filepath.Join(c.TmpMnt, "cpu")
+		// and the left side is taken exactly as written. Further, recall that in bind mounts, the
+		// remote side is the "device", and the local side is the "target."
```
brho, line 303: i always find this remote vs local confusing.  i guess "device" is the machine with the keyboard and terminal, where cpu runs?

client/fns_test.go
```diff
@@ -34,3 +34,50 @@ func TestVsockIdPort(t *testing.T) {
 		}
 	}
 }
+
+func TestParseBinds(t *testing.T) {
+	for _, tt := range []struct {
+		namespace string
+		fstab     string
```
brho: minor, but it might be easier to read the tests if fstab was last?

client/client.go
rminnich: I figure Dial is safe, it basically means "time to talk to cpud" and that's a good time to assemble the environment variables.

client/fns.go
rminnich: I thought about changing this, but I decided to leave it, as it kind of points back to the reason.

client/fns.go
rminnich: it's confusing. Are there better words? We could even use the words client and cpud if that would be better?

client/fns_test.go
rminnich: yes. It would. Will fix.

client/fns.go
brho: not sure tbh.  =)

brho (approved):

PR #151, opened 2022-10-01
Refactor dns-sd into its own module and cmds

ericvh:
This patch refactors dns-sd to seperate it a bit more from mainline cpu.

The dns-sd code is now in a ds module.  There is now a dcpu and dcpud command line which incorporate the dns-sd functionality.

I've also added a docker-based integration test in TESTDOCKERCPU which tests both the original cpu/cpud and tests dcpu/dcpud.


ericvh:
@rminnich asked to revert previous dnssd patch sets and resubmit clean.

PR #155, opened 2022-10-04
Add dnssd support to cpu (#6)

ericvh:
Add a new module (ds) which can be used to register and discover cpu servers with dnssd.  Create seperate cmds for cpu and cpud (decpu and decpud) which incorporate the use of dnssd while still using same base client and server modules as original.

ds uses meta-data in the dns TXT region to relate information about the host decpud is running on (including GOARCH, GOOS, as well as core count, memory information, load information, and number of cpu clients currently connected.  This information can be augmented from the decpud command line to include other meta-data (such as near-NIC, near-GW, near-storage, near-GPU, near-robot, etc.)

Lookup is specified as a URI (with sensible defaults applied) which is used in place of a hostname and can be omitted to take advantage of defaults.  meta-data requirements and sorting can be specified as query parameters to the URI.  e.g. dnsrd://?arch=arm64&sort=load


rminnich (approved):
not a fan of the name but it's no worse than the name cpu, so ... I don't know why I can't have a rebase button but ... you'll have to update it and git push -f ...

ericvh:
> not a fan of the name but it's no worse than the name cpu, so ... I don't know why I can't have a rebase button but ... you'll have to update it and git push -f ...

rebased.

PR #156, opened 2022-10-04
Add go, gox, and docker build targets

ericvh:
This adds build rules for things other than rust to Makefile.

No dependency tracking - sorry.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,74 @@
-all:
-	cargo build --release
+ifeq ($(OS),Windows_NT)
+    CCFLAGS += -D WIN32
+    ifeq ($(PROCESSOR_ARCHITEW6432),AMD64)
+        CCFLAGS += -D AMD64
+    else
+        ifeq ($(PROCESSOR_ARCHITECTURE),AMD64)
+            CCFLAGS += -D AMD64
+        endif
+        ifeq ($(PROCESSOR_ARCHITECTURE),x86)
+            CCFLAGS += -D IA32
+        endif
+    endif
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,5 +1,74 @@
-all:
-	cargo build --release
+ifeq ($(OS),Windows_NT)
```
rminnich: does this actually work?

Makefile
```diff
@@ -1,5 +1,74 @@
...
+        endif
+    endif
+else
+    UNAME_S := $(shell uname -s)
+    ifeq ($(UNAME_S),Linux)
+        OS = Linux
+    endif
+    ifeq ($(UNAME_S),Darwin)
+		OS = Darwin
+    endif
```
rminnich: don't we need an error if nothing matches?

Makefile
ericvh: I haven't tested on Windows, but seems to work on Mac / Linux.

Makefile
ericvh: Well, if OS doesn't end up equaling Linux then go won't try and build cpud (because it fails on Darwin).  I could simplify the boilerplate there (that I shamelessly copied from stack overflow), but figured it might come in handy in the future.

ericvh: I simplified the OS detecetion in the most recent patch.

Makefile
ericvh: Simplified.

rminnich (approved):

PR #157, opened 2022-10-06
Update documentation for docker

ericvh:
Point people are the ghcr.io container which is now automatically
populated by gitlab CI.  The container version is pegged to the branch
at the moment so its cpu:main, I wasn't sure how you wanted to handle
release flow of u-root/cpu:latest, probably best for now to create a
custom action that is only triggerable by an owner to tag whatever is in
cpu:main to cpu:latest.  Once we've done that we can update docs again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,16 +69,11 @@ or if you have installed a version of docker buildx, you can build a multi-arch 
 % docker buildx build --platform "linux/amd64,linux/arm64,linux/arm/v7" --progress plain --pull -t "${USER}/cpu:latest" .
 ```
 
-There is a pre-built version of a multi-arch container here which you can substitute for the next section if you want to try out on arm (either arm64 or arm/v7): 
-```
-ericvh/cpu:latest
-```
-
-### Pre-build Docker container for trying out cpu (on x86 for now)
+### Pre-built Docker container for trying out cpu (on arm64 & amd64 for now)
 
 We have created a docker container so you can try cpu client and server:
 ```
-rminnich/cpu:latest
+ghcr.io/u-root/cpu:main
 ```
 
 It includes both the cpud (server) and cpu (client) commands. In the
@@ -109,21 +104,14 @@ That is how we avoid
 storing keys in the container itself.
 ```
 docker network create cpud
-# If you ran docker before, you need to remove the
-# old identity.
-docker rm cpud_test
-docker run --mount type=bind,source=$KEY.pub,target=/key.pub --mount type=bind,source=$KEY,target=/key --name cpud_test --privileged=true -t -i -p 17010:17010 rminnich/cpu:latest
-```
-
-Note: once the container is done, if you want to run it again with the
-same named, cpud_test, you must:
-```
-docker rm cpud_test
+# If you ran docker before and it failed in some way, you may need to remove the
+# old identity (e.g. docker rm cpud_test)
+docker run --rm -v $KEY.pub,target=/key.pub -v $KEY,target=/key -v /tmp:/tmp --name cpud_test --privileged=true -t -i -p 17010:17010 ghcr.io/u-root/cpu:main
 ```
 
 Then you can try running a command or two:
 ```
-docker exec -it  cpud_test  /bin/cpu -key /key localhost /bin/date
+docker exec -it -e PWD=/ cpud_test  /bin/cpu -key /key localhost /bin/date
 ```
 
 Remember, this cpu command is running in the container. You need to use the name /key in the
@@ -131,7 +119,7 @@ container, not $KEY.
 
 To see the mounts:
 ```
-docker exec -it  cpud_test  /bin/cpu -key /key localhost /bin/cat /proc/mounts
+docker exec -it -e PWD=/ cpud_test  /bin/cpu -key /key localhost /bin/cat /proc/mounts
 ```
 
 You might want to just get a cpu command to let you talk to the docker cpud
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,29 +104,22 @@ That is how we avoid
...
-docker rm cpud_test
+# If you ran docker before and it failed in some way, you may need to remove the
+# old identity (e.g. docker rm cpud_test)
+docker run --rm -v $KEY.pub,target=/key.pub -v $KEY,target=/key -v /tmp:/tmp --name cpud_test --privileged=true -t -i -p 17010:17010 ghcr.io/u-root/cpu:main
 ```
 
 Then you can try running a command or two:
 ```
-docker exec -it  cpud_test  /bin/cpu -key /key localhost /bin/date
+docker exec -it -e PWD=/ cpud_test  /bin/cpu -key /key localhost /bin/date
```
rminnich: why do I need the PWD=/ ? I never needed to do that before.

ericvh: This is something I experienced on mac when using the docker container to run the client.  PWD is probably set to /Users/erivan01/src/cpu at this point, which doesn't exist inside the docker - and for some reason the fact it can't get to PWD is treated as fatal.  I suppose the other thing I could do is start the cpud with /Users in its namespace.  You do a variation of this in your cpu embedded in docker example, but the initial example doesn't have this.

```
(base) FV7GG9FTHL:cpu erivan01$ docker exec -e PWD=/ decpud /bin/decpu $DEBUG -key /key . /bin/date

Sun Oct  9 21:30:10 UTC 2022
(base) FV7GG9FTHL:cpu erivan01$ docker exec decpud /bin/decpu $DEBUG -key /key . /bin/date

2022/10/09 21:30:19 CPUD: your $PWD  is not in the remote namespace
2022/10/09 21:30:19 CPUD(as remote):file does not exist
2022/10/09 21:30:19 Wait: Process exited with status 1
(base) FV7GG9FTHL:cpu erivan01$ 
``` 

This happens the same for cpu/cpud -- but the above example is decpu/decpud.  BTW, I have a fix for mac docker desktop and being able to use the dns-sd versions.  I've written a dns-sd relay that tunnels the multicast into the docker container/network (which is broken on the mac) -- I'm writing up docs now and will post to slack, but the code is here:  github.com/decent-e/dnssd (and there is a package to run to perform the docker side of the relay here: docker pull ghcr.io/decent-e/dnssd:main )

ericvh: PWD is not one of the environment variables set by docker by default: (at least not on OSX Docker Desktop):

(base) FV7GG9FTHL:cpu erivan01$ docker exec -i -t cpud /bin/printenv
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=589b5cf79160
TERM=xterm
HOME=/

(base) FV7GG9FTHL:cpu erivan01$ docker exec -i -t -e PWD cpud /bin/printenv
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=589b5cf79160
TERM=xterm
PWD=/Users/erivan01/src/cpu
HOME=/

I'll fix the tests appropriately -- but there is a meta issue obvious here about needing to do something different on OSX (/Users isn't in the docker namespace)

ericvh: I guess part of the way we can deal with this is not make lack of $PWD in namespace fatal?  I mean it might be, but it also might not be, so we should try to run regardless, right?  Although I suppose if we are doing something nasty (rm -rf) then maybe we only want to do that in the right $PWD.  hmmmm.....  the other thing to do is just fix default namespace on OSX

ericvh:
Realizing there were some bigger issues in my update, but there are also serious issues with running cpu inside container....

README.md
ericvh: so part of the problem here is when running cpu inside the container like the examples from the original documentation -- you are never going to have the paths from the original system (unless you are in /bin, /lib, /lib64 or /root).  We could include instructions for mapping in /home, but what if you are elsewhere?  Its almost like you have to volume mount / to /mnt/local, bind the cpu/cpud binaries to /mnt/local/bbin and then /chroot /mnt/local before executing cpu.  @rminnich - what am I missing?

ericvh: Okay, updated the docs a bit with an explanation why things are different when running the cpu client inside the docker container and added some notes about what to expect when running on OSX.

README.md
```diff
@@ -109,29 +104,27 @@ That is how we avoid
...
 ```
 
-Note: once the container is done, if you want to run it again with the
-same named, cpud_test, you must:
-```
-docker rm cpud_test
-```
+Then you can try running a command or two by using the embedded cpu client in the docker container.  
+_NOTE_: when you run cpu in this way, it does not immediately have access to your host's file system, 
+which means you won't be able to really leverage the back-mount and you'll hae to artificially set
```
rminnich: hae -> have
unless you're jmk

README.md
```diff
@@ -186,7 +191,7 @@ Host h
 Note that the cpu command is itself a 9p server; i.e., your instance of cpu runs your server. The remote
 cpu server may run as root, but all file accesses happen locally as you. Hence,
 the cpu command does not grant greater access to the local machine than you already possess.
-I.e., there is no privilege escalation.
+i.e., there is no privilege escalation.
```
rminnich: you sure about this? i.e. is starting a new sentence.

ericvh: I don't even remember making that change, I wonder if vscode was being clever....

README.md
ericvh: fixed

README.md
ericvh: fixed

rminnich (approved):

PR #160, opened 2022-10-15
Make MSIZE 64K, not 8K

rminnich:
Measurement between systems shows that this provides good performance. Growing the packet beyond 64K shows no additional improvement.


rminnich:
we should probably at some point make this an option but at that same time, we already have a lot of options ... ideally we could make this an option you control from the cpu command.

ericvh (approved):
Looks right to me.

PR #164, opened 2022-11-07
client: escape command arguments

Lencerf:
Currently, If we run
  `./cpu <host-name> ./binary --arg1 "value with space"`
The client wiil actually run the following command on the remote host
  `./binary --arg1 value with space`
where `arg1=value` and the binary gets two extra arguments `with` and `space`, because the quote is lost.

This PR imports the shellescape package. With the fix, the remote host will receive a correctly quoted command line:
  `./binary --arg1 'value with space'`

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -16,6 +16,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/alessio/shellescape"
 	"github.com/hashicorp/go-multierror"
 	"github.com/mdlayher/vsock"
 	"github.com/u-root/u-root/pkg/termios"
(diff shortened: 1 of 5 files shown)


brho (commented):
Regarding this from the commit message:

> With the fix, the remote
host will receive a correctly quoted command line:
  ./binary --arg1 'value with space'

does the remote machine get 'value with space' or "value with space"?  i.e. does it get single quotes or the original double quotes?

Lencerf:
@brho it gets single quotes.

brho (approved):

client/client.go
```diff
@@ -462,21 +463,8 @@ func (c *Cmd) Start() error {
...
-	// cpu, unlike ssh, does not pass the arguments to a shell.
-	// Unlike Plan 9 shells, Linux shells do gargantuan amounts
-	// of file IO for each command, and it's a very noticeable
-	// performance hit.
-	// TODO:
-	// Possibly the correct thing here is to loop over
-	// c.Args and print each argument as %q.
-	cmd += fmt.Sprintf(" %v", strings.Join(c.Args, " "))
+
+	cmd += " " + shellescape.QuoteCommand(c.Args)
```
rminnich, line 467: for i := range c.Args {
c.Args[i] = strings.ReplaceAll(c.Args[i], "'", "'\"'\"'") + "'"
}

I think is all you need.

rminnich: actually ... 
c.Args = QuoteArgs(c.Args)
put that for loop in that function
then add tests for that function

PR #166, opened 2022-11-17
client.go: Support Firecracker virtio-vsock

Lencerf:
Add a new network type `unix-vsock` and implement the Firecracker virtio-vsock model:
https://github.com/firecracker-microvm/firecracker/blob/main/docs/vsock.md#host-initiated-connections

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -5,6 +5,7 @@
 package client
 
 import (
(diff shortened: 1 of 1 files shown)


rminnich (approved):
this is why we can't have nice things.

PR #167, opened 2022-11-23
Fix verbose priting

Lencerf:
This PR fix the usage of verbose() and v(). v() will be the log function provided by the user and verbose() is used in the package. v() is not called directly besides verbose().

We need to consider 3 kinds of verbose printings:
1. printings from cpud ssh server
2. printings from cpud remote mode
3. printings from local cpu command

The 1st will be logged to server, so we do not need to care about terminal interface. For the 2nd and 3rd cases, logs will be printed or forwarded to local terminal window. Since local terminal will also be used as the ssh terminal window, we need to add the extra `\r\n` to the end of log messages.

Function v() of package session was never set when cpud is running in remote mode.

This PR sets up the v() function of package session at the beginning of CPUD process.

source file in `cmds/decpud` and `ds` are not changed yet.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -35,8 +35,9 @@ const (
 	DefaultNameSpace = "/lib:/lib64:/usr:/bin:/etc:/home"
 )
 
-// V allows debug printing.
-var V = func(string, ...interface{}) {}
+// v allows debug printing.
+// Do not call it directly, call verbose instead.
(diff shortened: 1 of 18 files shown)


Lencerf:
I will create smaller PRs to fix logs package by package. The first one: https://github.com/u-root/cpu/pull/170

PR #168, opened 2022-11-28
cpud: add a switch to 'register' cpud with a remote host

rminnich:
In VMs and IoT, it is useful to have some idea that a host has come up. Pinging the cpud port until cpud comes up does not work well, especially in systems using vsock. What is most desired is that cpud 'check in', or register, with a control host so that the contol host can keep track of which cpuds are up and working.

Once the cpud has registered, the control host can periodically run a command to ensure the host is still up.

This PR adds a flag (default off) which tells cpud to register, what address to use, and a second flag to tell cpud how long to wait for the register to succeed.

The register flag is a Go-style addr (host:port) which specifies a host to connect to once cpud has called Listen and before it calls Serve. cpud will use the network on which it listens. Assuming the connect succeeds, cpud will send the two-byte string 'ok' to indicate its status.

registerTO specifies the timeout in seconds for the connect.

Currently, a failure to register or write 'ok' does not block cpud from continuing on to accept connections. Further, cpud will only print an error message if debug is enabled.

This PR includes a test for the new function.

### DIFF
--- a/cmds/cpud/main.go
+++ b/cmds/cpud/main.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"log"
 	"os"
+	"time"
 
 	// We use this ssh because it implements port redirection.
 	// It can not, however, unpack password-protected keys yet.
@@ -28,6 +29,10 @@ var (
 	network   = flag.String("net", "tcp", "network to use")
 	port9p    = flag.String("port9p", "", "port9p # on remote machine for 9p mount")
 	klog      = flag.Bool("klog", false, "Log cpud messages in kernel log, not stdout")
+
+	// Some networks are not well behaved, and for them we implement registration.
+	registerAddr = flag.String("register", "", "address and port to register with after listen on cpu server port")
+	registerTO   = flag.Duration("registerTO", time.Duration(5*time.Second), "time.Duration for Dial address for registering")
 
 	pid1 bool
 )
--- a/cmds/cpud/serve.go
+++ b/cmds/cpud/serve.go
@@ -5,6 +5,7 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"math"
 	"net"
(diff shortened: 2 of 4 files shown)


cmds/cpud/serve.go
```diff
@@ -104,6 +105,32 @@ func listen(network, port string) (net.Listener, error) {
...
 
+func register(network, addr string, timeout int64) error {
+	if len(addr) == 0 {
+		return nil
+	}
+	// This is a bit of a hack for now, to test the idea.
+	// if registeraddr is not empty, Dial it over the network,
+	// and send the string "ok".
+	// This may fail because the host may have incorrectly requested a registration
+	// but may not be listening, OR may just be looking for a cheap with to set a
```
brho: s/cheap with/cheap way/  ?

cmds/cpud/main.go
```diff
@@ -29,6 +29,10 @@ var (
 	port9p    = flag.String("port9p", "", "port9p # on remote machine for 9p mount")
 	klog      = flag.Bool("klog", false, "Log cpud messages in kernel log, not stdout")
 
+	// Some networks are not well behaved, and for them we implement registration.
+	registerAddr = flag.String("register", "", "address and port to register with after listen on cpu server port")
+	registerTO   = flag.Int64("registerTO", 5, "timeout in seconds for Dial address for registering")
```
brho: is there a flag.Duration?

cmds/cpud/serve.go
```diff
@@ -104,6 +105,32 @@ func listen(network, port string) (net.Listener, error) {
...
+	// delay between listen and accept for networks that are not well behaved.
+	c, err := net.DialTimeout(network, addr, time.Second*time.Duration(timeout))
+	// N.B.: it fails to connect, it fails to connect.
+	if err != nil {
+		return err
+	} else {
+		// if it works, it works. It not, log it and move on.
```
brho: s/ It not / If not /

cmds/cpud/serve.go
```diff
@@ -104,6 +105,32 @@ func listen(network, port string) (net.Listener, error) {
...
+		if _, err := c.Write([]byte("ok")); err != nil {
+			return fmt.Errorf("Writing OK to register address: %w", err)
+		}
+	}
+	if err := c.Close(); err != nil {
```
brho: do you want to defer c.Close() up above, in case the c.Write failed and you returned early?

cmds/cpud/main.go
rminnich: ah,right, I keep forgetting that one. Fixed.

cmds/cpud/serve.go
rminnich: thanks, yep, done!

cmds/cpud/serve.go
rminnich: done

brho (approved):
awesome, thanks!

PR #169, opened 2022-11-29
Fix support for specifying dnssd instance

ericvh:
cpu dnssd URI didn't support specifying an instance. Augment parsing and lookup code to allow for specifying specific instance by passing full qualified dnssd path.

Bug originally reported by @rminnich 

### DIFF
--- a/ds/ds.go
+++ b/ds/ds.go
@@ -125,17 +125,21 @@ func Parse(uri string) (dsQuery, error) {
 		return result, fmt.Errorf("Not an dns-sd URI")
 	}
 
-	// following dns-sd URI conventions from CUPS dnssd://instance._type._proto.domain/?query
-	if u.Hostname() != "" {
+	// following dns-sd URI conventions from CUPS
+	//   (e.g. dnssd://instance._type._proto.domain/?query)
+	//   We are going to look for the _type._proto tuple and split the
+	//   components around it because domain could have more than one
+	//   . speperated coponent (it could be local or example.com)
+	if len(u.Hostname()) != 0 {
 		parts := strings.Split(u.Hostname(), ".")
 		found := false
 		for p, v := range parts {
-			if v[0] == '_' {
+			if strings.HasPrefix(v, "_") {
 				if p > 0 {
 					result.Instance = strings.Join(parts[0:p], ".")
 				}
 				if len(parts) > p+1 {
-					if parts[p+1][0] == '_' {
+					if strings.HasPrefix(parts[p+1], "_") {
 						result.Type = parts[p] + "." + parts[p+1]
 					}
 					p = p + 2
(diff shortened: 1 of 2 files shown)


ds/ds.go
```diff
@@ -124,13 +125,35 @@ func Parse(uri string) (dsQuery, error) {
...
 
-	// following dns-sd URI conventions from CUPS
-	if u.Host != "" {
-		result.Domain = u.Host
-	}
-	if u.Path != "" {
-		// remove prefix slash
-		result.Type = u.Path[1:]
+	// following dns-sd URI conventions from CUPS dnssd://instance._type._proto.domain/?query
+	if u.Hostname() != "" {
```
rminnich: the convention nowadays seems to be
if len(u.Hostname) == 0
I forget why.

ds/ds.go
```diff
@@ -124,13 +125,35 @@ func Parse(uri string) (dsQuery, error) {
...
+		parts := strings.Split(u.Hostname(), ".")
```
rminnich, line 134: it would be best if this were a function that came with tests. I am bad at this kind of string foo on a good day, and I am sure this is right, but ... phew. Anytime I see indexes being manipulated this way I worry.

ds/ds.go
```diff
@@ -124,13 +125,35 @@ func Parse(uri string) (dsQuery, error) {
...
+		found := false
+		for p, v := range parts {
+			if v[0] == '_' {
```
rminnich: Using functions like strings.Prefix is just more robust. Yes, I know this is right now, but what if Future Ron inserts some bug that causes v to be weird?

ds/ds.go
```diff
@@ -124,13 +125,35 @@ func Parse(uri string) (dsQuery, error) {
...
+				if p > 0 {
+					result.Instance = strings.Join(parts[0:p], ".")
+				}
+				if len(parts) > p+1 {
```
rminnich, line 141: Some sort of comment about what is going on here would be helpful for future readers maybe.

ds/ds.go
ericvh: yeah, I spent an hour in go sandbox testing, but those should be easy to move over and add tests for.

rminnich (approved):

ds/ds.go
```diff
@@ -125,12 +126,38 @@ func Parse(uri string) (dsQuery, error) {
 	}
 
 	// following dns-sd URI conventions from CUPS
-	if u.Host != "" {
-		result.Domain = u.Host
-	}
-	if u.Path != "" {
-		// remove prefix slash
-		result.Type = u.Path[1:]
+	//   (e.g. dnssd://instance._type._proto.domain/?query)
```
rminnich: Function comment should be of form
// Parse parses DNS-SD URI etc. etc. 

Hey, it ain't me, it's the various linters :-)

if you can make that tweak we are good to go.

ericvh: Done

PR #177, opened 2022-12-04
Multids

rminnich:
You can now easily run more than one command with a single decpu command, e.g.
./decpu -n 2 -key ~/.ssh/cpu_rsa -sp 17020 . date
or
./decpu -n 2 -key ~/.ssh/cpu_rsa -sp 17020 dnssd://?arch=amd64  date
Sat Aug  4 10:29:11 AM UTC 2012
Thu Jun 21 01:28:22 PM UTC 2012

./decpu -n 2 -key ~/.ssh/cpu_rsa -sp 17020 dnssd://?arch=amd64 hostname
atomicpi
api2

### DIFF
--- a/cmds/decpu/decpu.go
+++ b/cmds/decpu/decpu.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 
 	// We use this ssh because it implements port redirection.
 	// It can not, however, unpack password-protected keys yet.
@@ -42,7 +43,8 @@ var (
 	keyFile     = flag.String("key", "", "key file")
 	namespace   = flag.String("namespace", "/lib:/lib64:/usr:/bin:/etc:/home", "Default namespace for the remote process -- set to none for none")
 	network     = flag.String("net", "", "network type to use. Defaults to whatever the cpu client defaults to")
-	port        = flag.String("sp", "", "cpu default port")
+	numCPUs     = flag.Int("n", 1, "number CPUs to run on")
+	sp          = flag.String("sp", "", "cpu default port")
 	root        = flag.String("root", "/", "9p root")
 	timeout9P   = flag.String("timeout9p", "100ms", "time to wait for the 9p mount to happen.")
 	ninep       = flag.Bool("9p", true, "Enable the 9p mount in the client")
@@ -140,13 +142,13 @@ func usage() {
 	log.Fatalf("Usage: cpu [options] host [shell command]:\n%v", b.String())
 }
 
-func newCPU(host string, args ...string) error {
+func newCPU(host, port string, args ...string) error {
 	// note that 9P is enabled if namespace is not empty OR if ninep is true
 	c := client.Command(host, args...)
 	if err := c.SetOptions(
 		client.WithPrivateKeyFile(*keyFile),
 		client.WithHostKeyFile(*hostKeyFile),
-		client.WithPort(*port),
+		client.WithPort(port),
 		client.WithRoot(*root),
 		client.WithNameSpace(*namespace),
 		client.With9P(*ninep),
@@ -175,11 +177,10 @@ func newCPU(host string, args ...string) error {
 }
 
 func main() {
-	var c []*ds.LookupResult
-
 	flags()
 	args := flag.Args()
 	host := ds.DsDefault
+	port := *sp
 	a := []string{}
 	if len(args) > 0 {
 		host = args[0]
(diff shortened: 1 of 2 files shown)


ds/ds.go
```diff
@@ -275,6 +276,56 @@ func Lookup(query dsQuery) (string, string, error) {
...
+// LookupResult is returned over a channel, and has either a result or an error.
+type LookupResult struct {
+	Entry *dnssd.BrowseEntry
+	Error error
+}
+
+// LookupChan returns a chan of *LookupResult and a cancel function.
+// The cancel function can be called when the desired number of results has
+// been received; or for any other reason in fact.
+func LookupChan(query dsQuery) (chan *LookupResult, context.CancelFunc) {
```
ericvh: Take a look at my resubmitted sort patch -- the original intended function was to sort the list of responses before returning.  While this was previously used to just pick the top one (which usually was the one with the lowest load because that was the default sort parameter) -- you probably still want to pick the top n from that list before getting a list of all the possible ones.  Logic could be here in Lookup or in decpu, I'm flexible on that, but I think we need to integrate sorting into whatever solution we end up with.

cmds/decpu/decpu.go
```diff
@@ -207,13 +209,23 @@ func main() {
...
-		}
+		go func() {
+			defer close(cpus)
+			c, cancel := ds.LookupChan(dq)
+			defer cancel()
+			var amt int
+			for e := range c {
+				amt++
+				if amt > *numCPUs {
+					return
```
ericvh: okay, so if we were to move sorting to the client, instead of instantiating the cpus on every loop, we wait till we hit n, sort the list and then instantiate.  I suppose the downside is with the code the way it is we dispatch as soon as the remote tells us and otherwise we are waiting for our timeout before dispatching.  We could conditionalize holding off on dispatching ONLY if we detect sort -- but we'd need parse to return a bool telling us whether or not there is a sort involved (by default there is though).

ericvh (commented):
Need to work out how to re-integrate sort which complicates this logic slightly. Otherwise looks good.

ds/ds.go
abrender: Just pointing out "Rethinking Classical Concurrency Patterns" by bcmills - (video: https://www.youtube.com/watch?v=5zXAHh5tJqQ slides:   https://drive.google.com/file/d/1nPdvhB0PutEJzdCq5ms6UI58dp50fcAN/view).    Exposing channels as part of the public API of this package is risky and ambiguous.  Whose job is it to close the channel - the caller or the callee **(explicitely: can you please document the expected behavior (who will close the channel) in the docstring? thanks!)?**

rminnich: for now, let's start with just matching for now, sorting can come later. 

The go way to do this would be to
func LookupChan(query dsQuery, opts ...func([]*LookupResult) error) {
..
get the host list as []*LookupResult
for o := range opts {
if err := o(l); err != nil {
return err
}
}

And those opts would bring in sorting or whatever else we need. That way, default is not sorted, but sorting can be added.

rminnich: done. Thanks.

The reason for the chan is that one might overlap getting lookup results and initiating the cpu connections. This is a common pattern in HPC. 

In this case, I doubt the lookup will take that long, so I'm okay with the changes.

ds/ds.go
```diff
@@ -357,6 +358,67 @@ func Lookup(query dsQuery) (string, string, error) {
...
+		if l.Error != nil {
+			err = l.Error
+			break
+		}
+		ret = append(ret, l)
+		if len(ret) >= n {
+			break
+		}
+	}
+	return ret, err
```
ericvh: Wouldn't we just need the dsSort before return like in standard Lookup() to get the sorted responses?

rminnich: The problem I have with that is -- sorted by what key? I'm a bit leery of embedding a sort in here without knowing what the user wants to sort on?

rminnich: anyway ... I view the sort as another PR if you don't mind, makes bisect better

rminnich: btw the idiomatic way to do this ... you'd change it as follows:
type Opt func([]*LookupResult) ([]*LookupResult, error)

func LookupN(query dsQuery, n int, opts ...Opt) ([]*LookupResult, error) {

right before the return:
for _, o := range opts {
   ret, err = o(ret)
   if err != nil {
    return ret, err
}

}

That lets you refine in any way you choose.
I'd rather do that than wire in a sort anyway.

ericvh:
the keys are specified in the URI and there are defaults based on tenants
and load average (this is the way it worked before lookupN) but I’ll submit
in a separate PR.

ds/ds.go
ericvh: So ... here's the thing.  dsSort was already in the code as part of Lookup, you are essentially removing the sort functionality by switching to LookupN as the default, but you don't even change the README to remove the sort options nor do you remove the sort tests in TESTDOCKERCPU -- so its actually you that is breaking the bisect.

rminnich: LookupN was not in the package before, and the sorts were deadcode.

I'll close this PR, go ahead and do what you think best, but as it is it's hard to get "more than 1" from a lookup.

ericvh:
It wasn't deadcode, it was called in regular Lookup here: https://github.com/u-root/cpu/blob/fa5cb43f3cd637f2060f6273bd42114e078e11a6/ds/ds.go#L351

But you don't need to close the PR, just change where the sort happens to the client.  Hold up and I'll send you a PR for your PR.

ericvh:
see [feature/dsLookupN](https://github.com/u-root/cpu/tree/feature/dsLookupN) for how to fix

rminnich:
by the power vested in me as the creator of this project, I am approving this PR.

PR #180, opened 2022-12-17
9p server: set iouint to 0 on open and creat

rminnich:
If the iounit is set to 0, the client (linux kernel) will use up to msize-sizeof(p9header) requests for read and write, leading to dramatically better performance.


ericvh (approved):
Looks good, this is probably the same neighborhood of code where you need to start setting Qid.version - but we can probably do that in a different pull request.

PR #182, opened 2022-12-18
p9cpu: Start a gRPC-based Rust implementation

Lencerf:
This project will contain 4 crates: the binaries for the client and server, the library, and a 9p server library.


p9cpu/crates/p9cpu/src/p9cpu.rs
```diff
@@ -0,0 +1,58 @@
...
+    Unix,
+}
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    #[arg(long, value_enum, default_value_t = Net::Tcp)]
+    net: Net,
+
+    #[arg(long, default_value_t = 11200)]
```
rminnich: 17010 is the IANA standard.

p9cpu/crates/p9cpu/src/p9cpu.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[arg(long, default_value = "")]
+    namespace: String,
+
+    #[arg(long, default_value_t = false)]
+    tty: bool,
+
+    #[arg(long)]
+    fs_tab: Option<String>,
+
+    #[arg(long, default_value = "/p9cputmp")]
```
rminnich: for interoperability with scripts that use cpu, I think this should be the same, always, as the standard cpu. Else path names will differ depending on which command people use. Also, if we're going to change cpu to use /cputmp by default, we should check to make sure people are ok with that. 
For now, since everyone is used to /tmp, I'd prefer to keep it as /tmp

but overall, I'd most prefer it if this version of cpu was a command, with a different name, that acted the same.

rminnich:
I see real value in this when we have private networks and don't care about encryption ... will be interested to see where this goes.

rminnich:
The make is not totally clean:

   Compiling cpu v0.1.0 (/home/rminnich/go/src/github.com/u-root/cpu)
warning: unused import: `futures::Future`
 --> src/client/mod.rs:7:5
  |
7 | use futures::Future;
  |     ^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `Read`
 --> src/client/mod.rs:8:15
  |
8 | use std::io::{Read, Write};
  |               ^^^^

warning: `cpu` (bin "cpu") generated 2 warnings
    Finished release [optimized] target(s) in 8.95s

p9cpu/crates/p9cpu/src/p9cpu.rs
Lencerf: Done.

p9cpu/crates/p9cpu/src/p9cpu.rs
Lencerf: Done. Changed to `/tmp`. I don't think using `/tmp` is a good idea, since people may also want to bring their local `/tmp` together with `/lib` `/home` to the remote machine. We can discuss this later.

Lencerf:
> The make is not totally clean:
> 
> Compiling cpu v0.1.0 (/home/rminnich/go/src/github.com/u-root/cpu)
> warning: unused import: futures::Future
> --> src/client/mod.rs:7:5
> |
> 7 | use futures::Future;
> | ^^^^^^^^^^^^^^^
> |
> = note: #[warn(unused_imports)] on by default
> 
> warning: unused import: Read
> --> src/client/mod.rs:8:15
> |
> 8 | use std::io::{Read, Write};
> | ^^^^.
> 
> warning: cpu (bin "cpu") generated 2 warnings
> Finished release [optimized] target(s) in 8.95s

These warnings are from the existing rust code (https://github.com/u-root/cpu/blob/main/src/client/mod.rs), not from `p9cpu`.

p9cpu/crates/p9cpu/src/p9cpu.rs
rminnich: I think you are right. The standard from plan 9 is /mnt/term, not sure if that's right either. What do you recommend?

The problem with a network file system mount to a top level directory can cause trouble, I am still not sure what's best here.

rminnich (approved):

PR #184, opened 2022-12-20
mount: allow bind-mounting a file to a file

Lencerf:
A user may want to bring `/etc` to a remote machine except `/etc/resolv.conf`, since it may contains `nameserver 127.0.0.1` if the user has used dnsmasq. In that case, the user can add the following line to fstab:
```
/tmp/local/etc/resolv.conf /etc/resolv.conf none defaults,bind 0 0
```
to use the original dns config on the remote machine.

### DIFF
--- a/mount/mount.go
+++ b/mount/mount.go
@@ -11,6 +11,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"path"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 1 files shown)


mount/mount.go
```diff
@@ -61,11 +62,35 @@ func mount(m mounter, fstab string) error {
...
+		if src, e := os.Stat(dev); e == nil && !src.IsDir() && flags&unix.MS_BIND != 0 {
+			// Source dev is a file and we are going to do a bind mount.
+			if target, e := os.Stat(where); e != nil {
+				// Destination does not exist, so we are going to create an empty file.
+				if e := os.MkdirAll(path.Dir(where), 0666); e != nil {
+					// Creation failed.
+					err = multierror.Append(err, fmt.Errorf("cannot create dir %s: %s", path.Dir(where), e))
+					continue
+				}
+				if f, e := os.Create(where); err != nil {
```
rminnich: I think just using os.WriteFile(where, []byte{}, 0666) or some standard mode would be easier. It's best not to ignore errors from close anyway ...

rminnich (approved):
one comment but it's up to you.

mount/mount.go
Lencerf: Done.

PR #188, opened 2022-12-31
Fix ds to return a slice of results

ericvh:
This patch changes Lookup to return a slice of up to n results, allowing the client to access more than one host based on sorted results from dns query.

### DIFF
--- a/ds/ds.go
+++ b/ds/ds.go
@@ -338,10 +338,6 @@ func Lookup(query dsQuery, n int) ([]*LookupResult, error) {
 		respCh <- nil
 	}()
 
-	if err != nil {
-		return nil, err
-	}
-
 	for {
 		e := <-respCh
 		if e == nil {
@@ -351,7 +347,7 @@ func Lookup(query dsQuery, n int) ([]*LookupResult, error) {
 	}
 
 	if len(responses) == 0 {
-		return nil, fmt.Errorf("dnssd found no suitable service")
+		return nil, fmt.Errorf("dnssd found no suitable service %w", err)
 	}
 
 	dsSort(query.Text, responses)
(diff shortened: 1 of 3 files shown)


ds/ds.go
```diff
@@ -333,7 +339,7 @@ func Lookup(query dsQuery) (string, string, error) {
 	}()
 
 	if err != nil {
-		return "", "", fmt.Errorf("dnssd failed: %w", err)
+		return nil, err
 	}
 
 	for {
```
rminnich, line 341: so the way that is common here ... 
reader of LookupResult
for e := range <- respCh {
responses = ...append(...)
}

no testing for nil: producer of responses is supposed to count the number needed and close the channel when done.

On the writer side, it simply writes to the chan, and closes it when done.

see this for an example.

https://go.dev/play/p/AhR__tFPcrT

I do several mean things there: make the chan too small for the write loop, and don't dequeue all the entries on the read side, it works out ok. The default in the select handles the case that the chan can't be written to.

this removes your nil test, and will ensure the chan is closed and we don't leak goroutines and so on.

so the reader takes what the writer sends.
on the writer side, it would run 

anyway, nemo had his own go dialect where readers could close chans, but that we do not have.


I dunno, do I need the select ?I thought I did but maybe not.

ericvh: So I was trying to do a minimal change to the Lookup code to get you your slice -- it is quite possible there was something wrong in the original Lookup code.  Looking at the flow, that error check is likely useless and probably needs to go away -- if there is an error, respChan will get a nil which will break the for loop and then we can check err on the other side if we like and report something more useful than "no suitable service" which would likely be the default.

Otherwise since we are spawning the goroutine with with a ctxt, I guess I figured the context timeout would clean up anything lingering....

ericvh: okay, I've removed the vestigial error in the patch update.

rminnich (approved):
I'm looking forward to using this!

PR #195, opened 2023-01-18
cpu: forward SIGINT/SIGTERM to remote commands

Lencerf:

### DIFF
--- a/cmds/cpu/cpu.go
+++ b/cmds/cpu/cpu.go
@@ -12,8 +12,10 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	// We use this ssh because it implements port redirection.
 	// It can not, however, unpack password-protected keys yet.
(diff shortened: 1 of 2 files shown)


cmds/cpu/cpu.go
```diff
@@ -149,17 +159,50 @@ func newCPU(host string, args ...string) error {
...
+	defer close(errChan)
+
+	go func() {
+		verbose("start")
+		if err := c.Start(); err != nil {
+			errChan <- fmt.Errorf("Start: %v", err)
+			return
+		}
+		verbose("wait")
+		if err := c.Wait(); err != nil {
```
rminnich: another way to do this:
errChan <- c.Wait()
has the same effect.

cmds/cpu/cpu.go
```diff
@@ -149,17 +159,50 @@ func newCPU(host string, args ...string) error {
...
+			var sigErr error
+			switch sig {
+			case syscall.SIGINT:
+				sigErr = c.Signal(ossh.SIGINT)
+			case syscall.SIGTERM:
+				sigErr = c.Signal(ossh.SIGTERM)
+			}
+			verbose("signal %v sent to %q", sig, c.Args[0])
+			if sigErr != nil {
+				err = fmt.Errorf("sending %v to %q: %v", sig, c.Args[0], sigErr)
```
rminnich: I would argue that we should not quit here, log an error and continue?

server/server.go
```diff
@@ -62,6 +62,23 @@ func handler(s ssh.Session) {
...
 	cmd := command(a[0], a[1:]...)
+
+	sigChan := make(chan ssh.Signal, 1)
+	defer close(sigChan)
+	s.Signals(sigChan)
+	go func() {
+		for signal := range sigChan {
+			switch signal {
+			case ssh.SIGTERM:
+				cmd.Process.Signal(syscall.SIGTERM)
```
rminnich: should we check for an error here?

Lencerf: Done

cmds/cpu/cpu.go
Lencerf: Done.

cmds/cpu/cpu.go
Lencerf: Done.

rminnich (approved):

PR #198, opened 2023-02-02
session:Use ip6 and ip4 loopback address, not "localhost"

rminnich:
We had moved to using the name "localhost", since at some sites, only ip6 or ip4 protocols are not provided, and, hence, using the name localhost allowed setting a preference.

cpu should be able to function with missing or damaged host files, however.

Change the Dial code in session to use the two well known ip loopback address, ::1, and 127.0.0.1, in that order.


ericvh (approved):
Can confirm that this does seem to solve the problem I was having (sorry it to so long to double check -- was a bit convoluted to get the patch into my build flow).

PR #210, opened 2023-04-22
cpud: remove CPU_NAMESPACE

Lencerf:
This commit follows 0de8d08 ("client: only pass CPU_FSTAB to cpud") and removes all CPU_NAMESPACE related codes from the cpud side.


rminnich (approved):
very nice, we needed to do this for a long time now.

PR #211, opened 2023-04-22
session: remove cpu related env

Lencerf:
CPU_FSTAB and CPU_TMPMNT are only used by cpud. Hide them from the actual command.


rminnich (commented):
I'm not so sure I like this, because in debugging I use these values. Is this fixing a known problem?

PR #212, opened 2023-04-23
p9cpu: implement rpc server methods

Lencerf:
Add a new dummy Server struct which will implement the actual cpu server logic.


p9cpu/crates/libp9cpu/src/server.rs
```diff
@@ -0,0 +1,73 @@
+use crate::cmd;
+use async_trait::async_trait;
+use futures::Stream;
+use thiserror::Error;
+use tokio::sync::mpsc;
+use tokio_stream::wrappers::ReceiverStream;
+
+#[async_trait]
+pub trait P9cpuServerT {
```
rminnich, line 9: Is the T suffix now a standard? It seems a bit awkward.

rminnich (approved):

rminnich (approved):

p9cpu/crates/libp9cpu/src/server.rs
Lencerf: No. I see some discussions but No standard yet https://github.com/rust-lang/api-guidelines/discussions/28.

PR #215, opened 2023-04-27
Client can not specify command

rminnich:
also includes a fix.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -62,7 +62,7 @@ func errval(err error) error {
 func handler(s ssh.Session) {
 	a := s.Command()
 	verbose("handler: cmd is %q", a)
-	cmd := command(a[0], a[1:]...)
+	cmd := command("cpud", append([]string{"-remote"}, a...)...)
 
 	sigChan := make(chan ssh.Signal, 1)
 	defer close(sigChan)
--- a/server/server.go
+++ b/server/server.go
@@ -59,10 +59,10 @@ func errval(err error) error {
 	return err
 }
 
-func handler(s ssh.Session) {
+func handler(s ssh.Session, cpud string) {
 	a := s.Command()
 	verbose("handler: cmd is %q", a)
-	cmd := command("cpud", append([]string{"-remote"}, a...)...)
+	cmd := command(cpud, append([]string{"-remote"}, a...)...)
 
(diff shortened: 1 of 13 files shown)


server/server.go
```diff
@@ -62,7 +62,7 @@ func errval(err error) error {
 func handler(s ssh.Session) {
 	a := s.Command()
 	verbose("handler: cmd is %q", a)
-	cmd := command(a[0], a[1:]...)
+	cmd := command("cpud", append([]string{"-remote"}, a...)...)
```
Lencerf: How about using `os.Args[0]` instead of `cpud`?

rminnich: Let me think on that. I am not sure. I can see reasons for both paths.

rminnich: e.g., what if we started cpud as /init, but we really want to run with argv[0] = cpud. but, yeah, I'm tempted to like this idea, but ... hmm, hmm, hmmm, hmmmmmmmm, not sure.

brho: from a debugging/hacking perspective, if i was trying something out and renamed my new version "cpud-new", but then it turns out it was calling the original cpud that was still in $PATH, i'd be gouging my eyeballs out.

ericvh: yeah - I was just trying out the debugging use case like the other day.

Lencerf: If we have `/init` linked to `/bin/cpud`, and start the cpud server as `/init`, when cpud starts the `-remote` subprocess, we will have `/init -remote <....>`, which will still work correctly.

rminnich: I think I'd like to stick with cpud as the name to exec instead of argv[0]. But I can't quite tell from comments what brho and eric think. This allows playing with PATH and namespaces to get a new cpud from the server. But it's definitely not a clear decision.

Lencerf: `server` is a package and both `cpud` and `decpud` use it. If someone uses `decpu` to talk to `decpud`, the `decpud` server will end up invoking `cpud` to launch a task.

rminnich: yeah. That's another concern.

rminnich:
ok, look at this. I was not comfortable having a package blindly use os.Args

Lencerf (approved):

PR #216, opened 2023-04-27
p9cpu: implememnt rpc-based cpu cli

Lencerf:


p9cpu/crates/libp9cpu/src/lib.rs
```diff
@@ -33,3 +35,40 @@ where
         }
     }
 }
+
+pub const NINEP_MOUNT: &str = "mnt9p";
+pub const LOCAL_ROOT_MOUNT: &str = "local";
+
+pub fn parse_namespace(namespace: &str, tmp_mnt: &str) -> Vec<cmd::FsTab> {
```
rminnich, line 42: do we need this any more?

p9cpu/crates/p9cpu/src/p9cpu.rs
```diff
@@ -40,7 +45,43 @@ struct Args {
...
+            args.port,
+        )),
+        Net::Unix => libp9cpu::Addr::Uds(args.host),
+        Net::Tcp => libp9cpu::Addr::Tcp(format!("{}:{}", args.host, args.port).parse()?),
+    };
+
+    if args.tmp_mnt.is_empty() {
+        bail!("`tmp_mnt` cannot be empty");
+    }
+    let mut fs_tab_lines = parse_namespace(&args.namespace, &args.tmp_mnt);
```
rminnich, line 60: I thought we were dropping namespace? Or am I missing something.

rminnich (approved):

p9cpu/crates/libp9cpu/src/lib.rs
Lencerf: This is the client side. We still provide cpu_namespace as a convenient way to specify an fstab. The client will convert the namespace to an fstab. The server side does not use namespaces.

PR #217, opened 2023-04-28
p9cpud: implement the server

Lencerf:
When a server receives a new command request, it listens on an unix domain socket, and at the same time, starts a new instance of self with flag `--launch <uds>`, the sub process connects to the uds, gets the command request, sets up the namespaces, and exec the actual command.

### DIFF
--- a/p9cpu/crates/libp9cpu/src/server.rs
+++ b/p9cpu/crates/libp9cpu/src/server.rs
@@ -1,63 +1,299 @@
+use crate::async_fd::AsyncFd;
 use crate::cmd;
 use async_trait::async_trait;
 use futures::Stream;
+use prost::Message;
+use std::fmt::{Debug, Display};
+use std::hash::Hash;
+use std::os::unix::prelude::{FromRawFd, OwnedFd};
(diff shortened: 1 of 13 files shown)


p9cpu/crates/libp9cpu/src/server.rs
```diff
@@ -1,63 +1,314 @@
...
+            return Err(Error::DuplicateId);
+        }
+        let mut handles = vec![];
+        let mut ninep_port = None;
+        if let Some((port, handle)) = ninep.write().await.take() {
+            ninep_port = Some(port);
+            handles.push(handle);
+        }
+
+        let uds_path = format!("/tmp/p9cpud-launcher-{}", &sid);
```
rminnich: is this in a private namespace?

rminnich (approved):

p9cpu/crates/libp9cpu/src/server.rs
Lencerf: In the new commit, I switched to a tcp port to avoid conflicts.

PR #230, opened 2023-08-04
client: add a union 9p server

rminnich:
The union9p server gathers provides a single 9p mount for several 9p servers.

It works much as the Plan 9 namespaces work. Callers specify a set of names. On Walk, the names are matched against the names in the Walk, and the first match is used for the Walk.

As in Plan 9, the entire Walk request is sent to the server once a match hits, not a portion.
i.e., if a the Walk is to "home", "rminnich", and there is an entry for
home
Then the matching server will receive "home", "rminnich", not just "rminnich"

### DIFF
--- a/client/union9p.go
+++ b/client/union9p.go
@@ -1,0 +1,396 @@
+// Copyright 2018 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


Lencerf (approved):

brho (commented):
regarding the comment:
```
i.e., if a the Walk is to "home", "rminnich", and there is an entry for
home Then the matching server will receive "home", "rminnich", not just
"rminnich"
```
does that mean if (in plan 9) i bind some server to /secret-path-name/, then do a walk /secret-path-name/server-file, that the server sees "secret-path-name"?  i though the server would only see a walk from its "/"

though i do agree that if you did a walk to /secret-path-name/multi/directory/walk, that the server would see all of "/multi/directory/walk".

PR #231, opened 2023-08-05
Let callers provide an p9.Attacher

rminnich:
This is preparing for serving from a flattened docker container

Also, export CPU9P, external users will need it.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -16,6 +16,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/hugelgupf/p9/p9"
 	"github.com/mdlayher/vsock"
 	"github.com/u-root/u-root/pkg/termios"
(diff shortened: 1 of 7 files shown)


rminnich:
This is for sidecore, which is working. It is the next in a series of changes. Almost there.

client/client.go
```diff
@@ -142,6 +144,16 @@ func Command(host string, args ...string) *Cmd {
 // Set is the type of function used to set options in SetOptions.
 type Set func(*Cmd) error
 
+// WithServer allows setting custom 9P servers.
+// One use: should users with to serve from a flattened
```
brho: s/with/wish/

brho (approved):

client/client.go
rminnich: fixed, thanks, PTAL.

PR #236, opened 2023-08-14
Upgrade p9, use templates to shrink code base a little, & run go mod tidy

hugelgupf:
Not doing this for unionfs as I'm getting ready to commit that upstream in https://github.com/hugelgupf/p9/pull/75 so we'll be able to replace that wholesale (hopefully)


rminnich (approved):
looks good. Thanks. I kind of wish we could we have
templatefs.XattrUnimplemented("cpio")
such that on errors we could see what server it was for. It can be very hard to debug sometimes. That said, this is really a big improvement.

PR #242, opened 2023-09-01
cpud: default to building cpud with mDNS enabled

rminnich:
This relies on a double negative:
// go:build !nomDNS

u-root is not properly handling
-tags mDNS
in the command line, and we kind of need this, so
I will leave it to Google to work around it.


rminnich:
hold off on this, still working to find out what's up.

I mean approve if you want but I don't commit yet.

rminnich:
not needed. Not sure what was going wrong.

PR #245, opened 2023-09-14
README: add notes on and demo of cpu over USB

orangecms:
still gotta add DT nodes like
```
&usb_otg {
	dr_mode = "peripheral";
	status = "okay";
};
```

see also https://github.com/orangecms/linux/blob/534d38db598042fce9ccaf35489e21caa523abb2/arch/riscv/boot/dts/allwinner/sun20i-d1-mangopi-gpi15.dts#L258

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,22 @@ machine.
 We can use the same trick to cpu to Linux from OSX, but instead of having an Arm tree under `pwd`
 we'll need a Linux binary tree to pick binaries from.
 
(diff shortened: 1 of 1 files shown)


rminnich (approved):
In the old Unix world and in Plan 9, you can tell you might have got it right when you try something like this and, without a single change to code, it "just works". 

This is fantastic. It is a perfect example of the spirit of Unix and Plan 9. Thank you.

PR #248, opened 2023-09-23
cpud: add an option to sleep before serving

rminnich:
When cpud runs as init, and serves mDNS, it can start serving before mDNS will work. It is still not clear why this fails, but adding a 5 second sleep before serve() resolves it for now. Such a sleep is obviously undesirable, but the default is 0.

To set the timeout when cpud runs as init, add -- --sleepBeforeServing=5s to the kernel command line.


orangecms (approved):
Always in favor of a good nap! :sleeping:

orangecms:
@rminnich I've enabled the auto-merge option in the settings. That label is obsolete and only worked with some 3rd party integration up to some point.

This is now an option of the merge button instead.

PR #250, opened 2023-10-11
cpu9p: implement p9.File.Lock

Lencerf:


rminnich (approved):
Is there any way to add a test? If not, that's ok, file locking is a bogus idea anyway :-)

Lencerf:
> Is there any way to add a test? If not, that's ok, file locking is a bogus idea anyway :-)

It's not easy because this lock is process-associated, meaning in the test process, it's hard to create a lock conflict, or I need to spawn a subprocess.

PR #253, opened 2023-11-05
Windows

rminnich:
yep, not kidding. This lets me work under windows. It's not all there, consider this a WIP until next week.


client/cpu9p_windows.go
```diff
@@ -0,0 +1,141 @@
+// Copyright 2018 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
rjoleary, line 3: You should mention Apache in the top-most LICENSE file -- some people might be surprised that some files are under different licenses.

client/cpu9p_windows.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"errors"
+	"os"
+
+	"github.com/hugelgupf/p9/p9"
+)
+
+// GetAttr implements p9.File.GetAttr.
+//
+// Not fully implemented.
```
rjoleary, line 26: Mention one thing that's not implemented. Or is it not implemented because of Windows?

client/cpu9p_windows.go
```diff
@@ -0,0 +1,141 @@
...
+		mode |= p9.ModeDirectory
+	}
+	attr := p9.Attr{
+		Mode:             mode,
+		UID:              0,
+		GID:              0,
+		NLink:            3,
+		Size:             uint64(fi.Size()),
+		BlockSize:        uint64(512),
+		Blocks:           uint64(fi.Size() / 512),
```
rjoleary, line 49: I feel like this should round up: `(size + 511) / 512`

client/cpu9p_windows.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	valid := p9.AttrMask{
+		Mode:   true,
+		UID:    true,
+		GID:    true,
+		NLink:  true,
+		RDev:   true,
+		Size:   true,
+		Blocks: true,
+		ATime:  true,
```
rjoleary, line 65: It seems like `ATime` stands for "Access Time" and we're setting it to `ModTime` which is "Modified Time" which seems incorrect. Is it possible to just set this AttrMask value to false to indicate we don't know the `ATime`?

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/gliderlabs/ssh v0.3.5
 	github.com/hugelgupf/p9 v0.2.1-0.20230814004337-e6037077d6dc
 	github.com/kevinburke/ssh_config v1.2.0
-	github.com/u-root/u-root v0.11.1-0.20230913033713-004977728a9d
+	github.com/u-root/u-root v1.0.1
```
rjoleary, line 9: Is this intentional? I thought we used Go-style versioning where it's risky to go above v0.x.

rjoleary (commented):
Windows support is a hug leap!

rminnich:
Thanks so much for that review! I will tidy things up tomorrow. Your suggestions are excellent.

rminnich:
actually, I can't move forward on this until the termios windows support on u-root is upstreamed, though I may fix that problem (the review is never happening?) and move to using some other termios package, as gosh seems to do.

client/cpu9p_windows.go
rminnich: done

rminnich:
idk about windows any more

PR #255, opened 2023-11-23
Move to golang x term

rminnich:
This works on darwin and linux. 

The golang x term package has less drama :-) but less capability than u-root termios (which we don't need or use, so who cares)

### DIFF
--- a/client/fns_test.go
+++ b/client/fns_test.go
@@ -6,7 +6,6 @@ package client
 
 import (
 	"errors"
-	"os"
 	"path"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


rminnich:
I will test this over the weekend, but this should make windows work too.

client/fns_test.go
```diff
@@ -36,29 +39,53 @@ func TestVsockIdPort(t *testing.T) {
 }
 
 func TestParseBinds(t *testing.T) {
+	td := func(s string) string {
+		return path.Join(os.TempDir(), s)
```
binjip978: maybe t.TempDir() will be idiomatic?

rminnich: both were wrong, it has to /tmp :-)
see comment as to why.

binjip978 (approved):

PR #258, opened 2023-12-08
Add a new way to create CPIO9P: NewCPIO9PReaderAt

rminnich:
TODO:
o find a way to use cpio.InMemArchive
o this should be a package


rminnich:
This is needed so sidecore can use compressed CPIO archives

binjip978 (approved):

PR #262, opened 2023-12-23
ds: fix staticcheck warnings

binjip978:


Lencerf (approved):

ds/ds.go
```diff
@@ -236,12 +236,12 @@ func genSortTxt(key string, operator byte) lessFunc {
 				return false
 			}
 		}
-		n1, err := strconv.ParseFloat(c1.Text[key], 10)
+		n1, err := strconv.ParseFloat(c1.Text[key], 64)
```
rminnich, line 239: why make this 64? I assume the author had a reason to limit to 10 bits? Same question applies to all parsefloat changes.

rminnich (changes requested):
Can you confirm there is a reason for the parsefloat changes? I have to assume ericvh chose 10 bits for a reason.

ds/ds.go
binjip978: we can make it 32, but this change really have no effect, it will still run atof64 like with 10 that we have right now
```go
func ParseFloat(s string, bitSize int) (float64, error) {
	f, n, err := parseFloatPrefix(s, bitSize)
	if n != len(s) && (err == nil || err.(*NumError).Err != ErrSyntax) {
		return 0, syntaxError(fnParseFloat, s)
	}
	return f, err
}

func parseFloatPrefix(s string, bitSize int) (float64, int, error) {
	if bitSize == 32 {
		f, n, err := atof32(s)
		return float64(f), n, err
	}
	return atof64(s)
}
```

SA1030 - Invalid argument in call to a strconv function
This check validates the format, number base and bit size arguments of the various parsing and formatting functions in strconv.

rminnich: I'm still not understanding. The bitSize argument is not ignored -- it is used in the call to parseFloatPrefix. The author set it to 10 in this call. What is the reason to change it to 64?

binjip978: So the reason is that only "valid" arguments is 32 or 64.

If bitSize is 32 it will parse to float32 and return float64 of a result, otherwise it will always use atof64. See bitSize only used in 

```
if bitSize == 32  { ...
}
```

So if we set bitSize to 10, -10 or 25 it will have the same effect as 64, but it confusing for the reader.

https://staticcheck.dev/docs/checks/#SA1030

Also from documentation:
ParseFloat converts the string s to a floating-point number with the precision specified by bitSize: 32 for float32, or 64 for float64. When bitSize=32, the result still has type float64, but it will be convertible to float32 without changing its value.

hugelgupf: https://pkg.go.dev/strconv#ParseFloat 10 is equivalent to 64, as it is the default. But 64 really does make more sense.

rminnich (approved):

rminnich:
thanks for your patience :-)

PR #266, opened 2024-03-11
Add flag -useKey for ssh without keys

Lencerf:
In some fully isolated environments, the key used to encrypt the ssh connection is redundant. With this commit, we can use cpud/cpu without a key file by
* server: ./cpud -pk=
* client: ./cpu -useKey=false $HOST bash

-useKey as default is True.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -48,23 +48,24 @@ type Cmd struct {
 	// be set directly.
 	Host string
 	// HostName as found in .ssh/config; set to Host if not found
-	HostName       string
-	Args           []string
-	Root           string
-	HostKeyFile    string
-	PrivateKeyFile string
-	Port           string
-	Timeout        time.Duration
-	Env            []string
-	SessionIn      io.WriteCloser
-	SessionOut     io.Reader
-	SessionErr     io.Reader
-	Stdin          io.Reader
-	Stdout         io.Writer
-	Stderr         io.Writer
-	Row            int
-	Col            int
-	hasTTY         bool // Set if we have a TTY
+	HostName          string
+	Args              []string
+	Root              string
+	HostKeyFile       string
+	PrivateKeyFile    string
+	DisablePrivateKey bool
+	Port              string
+	Timeout           time.Duration
+	Env               []string
+	SessionIn         io.WriteCloser
+	SessionOut        io.Reader
+	SessionErr        io.Reader
+	Stdin             io.Reader
+	Stdout            io.Writer
+	Stderr            io.Writer
+	Row               int
+	Col               int
+	hasTTY            bool // Set if we have a TTY
 	// NameSpace is a string as defined in the cpu documentation.
 	NameSpace string
 	// FSTab is an fstab(5)-format string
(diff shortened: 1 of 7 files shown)


client/fns.go
```diff
@@ -67,11 +67,8 @@ func (n nonce) String() string {
 func (c *Cmd) UserKeyConfig() error {
 	kf := c.PrivateKeyFile
 	if len(kf) == 0 {
-		kf = config.Get(c.Host, "IdentityFile")
```
rminnich, line 70: This breaks all the ssh support from what I can see. Did I miss something?

Lencerf: I do not think so. Function `getKeyFile` in cmds/cpu/cpu.go has handled the case where `kf` is en empty string
https://github.com/u-root/cpu/blob/e30097ae68b3c1581d5a8ed1c65c02b5382e358f/cmds/cpu/cpu.go#L82-L98

Basically, before this change, the case of an empty `kf` is handled twice: in `getKeyFile` of cmds/cpu/cpu.go and `UserKeyConfig` of client/fns.go. `UserKeyConfig` is only called in the `Dial` method of `Cmd`. Therefore, if we use the binary `cpu` command, the code path of `if len(kf) == 0` in `UserKeyConfig` is never executed.

After this change, if a user has flag `-key="" -useKey=true`,  `getKeyFile` in `cpu.go` will call `config.Get(c.Host, "IdentityFile")`. On the other hand, the library function `UserKeyConfig` always think that `c.PrivateKeyFile` is the key file a user want to use. If it is empty, that means the user does not want to use a key.

Lencerf: This change should not break any ssh use case if a user uses the `cpu` command binary. 

However, if a user used the `client` library and `c.PrivateKeyFile` is empty, yes the behavior is changed,
* before: `UserKeyConfig` will call `config.Get(c.Host, "IdentityFile")` to find a default key
* after: `UserKeyConfig` thinks that the user does not want to use a key.

If this kind of change is unwanted, what about adding an additional field `useKey` in the `Cmd` struct?

rminnich: sidecore uses that package and I'll need to see how this affects it. I think we should just add a useKey flag instead of counting on behavior when the string is zero length. My bad.

Lencerf: I added a `DisablePrivateKey` field in the `Cmd` struct, and it is `false` as default. Then the behavior of the `client` package is unchanged.

client/client.go
```diff
@@ -48,23 +48,24 @@ type Cmd struct {
...
-	Stderr         io.Writer
-	Row            int
-	Col            int
-	hasTTY         bool // Set if we have a TTY
+	HostName          string
+	Args              []string
+	Root              string
+	HostKeyFile       string
+	PrivateKeyFile    string
+	DisablePrivateKey bool
```
rminnich, line 56: Very nice. It means the "zero value" is the safer value -- use a private key.

client/client.go
```diff
@@ -218,6 +219,14 @@ func WithPrivateKeyFile(key string) Set {
 	}
 }
 
+// WithPrivateKeyFile disables using private keys to encrypt the SSH connection.
```
rminnich: The comment name does not match the function name.

rminnich (commented):
I have not tested this, I'm assuming you have tested it several ways, and is there a test written?

client/client.go
Lencerf: ops! fixed now.

Lencerf:
> I have not tested this, I'm assuming you have tested it several ways, and is there a test written?

I added a few unit tests. I manually tested with QEMU.

rminnich (approved):

rminnich:
very nice, we've needed this for a long time.

PR #275, opened 2024-05-16
cpu: set TMPDIR environment variable to /tmp

rminnich:
Set the TMPDIR environment variable to /tmp, always.


rminnich:
This allows the cpu command to run from OSX

Lencerf (approved):

PR #49, opened 2021-12-17
go.mod: Update go.mod to use correct versioning of u-root

walterchris:
u-root versioning changed and versions vX.0.0 are deprecated now. This
commits changes the version used within go.mod from v7.0.0 to v0.8.1.

Signed-off-by: Christian Walter <christian.walter@9elements.com>

PR #276, opened 2024-05-18
bump billy to use a pseudo-release of main

rminnich:
I fixed a bug in billy:
boundos:insideBaseDirEval: return true if baseDir is "/" #49

This really requires a new release, but they are not ready to cut oen: Per the billy maintainer:
"Hey @rminnich, I would recommend bumping to the latest pseudo version from master as I am not yet sure when we will cut a new release."

use a pseudo version for billy


rminnich:
With this change, I can now use --nfs=true on osx

The cpio stuff is not working yet, but it's close.

Lencerf (approved):

PR #279, opened 2024-07-10
Cpu namespace over ssh

rminnich:
Did you want to run cpu, but don't want to set up cpud?
If you have an sshd, you're good to gol

This is cpu talking to an sshd.
```
rminnich@pop-os:~/go/src/github.com/u-root/cpu/cmds/cpu$ cpu -nfs=true -9p=false  q date
Wed Jul 10 03:08:16 PM PDT 2024
```

This is cpu talking to an sshd, but setting up the namespace.
```
rminnich@pop-os:~/go/src/github.com/u-root/cpu/cmds/cpu$ cpu -nfs=true -9p=false  q ~/bin/ns
[sudo] password for rminnich:
rminnich@pop-os:~$ mount
...
127.0.0.1:9973ecb0-b836-4f7b-997d-d7bc7dc08456 on /tmp/cpu type nfs (rw,relatime,vers=3,rsize=1048576,wsize=1048576,namlen=255,hard,nolock,proto=tcp,port=44957,timeo=600,retrans=2,sec=sys,mountaddr=127.0.0.1,mountvers=3,mountport=44957,mountproto=tcp,local_lock=all,addr=127.0.0.1)
127.0.0.1:9973ecb0-b836-4f7b-997d-d7bc7dc08456/usr/lib on /usr/lib type nfs (rw,relatime,vers=3,rsize=1048576,wsize=1048576,namlen=255,hard,nolock,proto=tcp,port=44957,timeo=600,retrans=2,sec=sys,mountaddr=127.0.0.1,mountvers=3,mountport=44957,mountproto=tcp,local_lock=all,addr=127.0.0.1)

...
```

ns is a script:
```
#!/bin/sh

sudo -E unshare -m ~rminnich/go/bin/cpuns
```

So the key new command here is cpuns.

So, if you want cpu, with just a plain old sshd, you've got it now.

There are a few issues left, but if you want early access, here it is.

I don't even know how to debug it ...

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -159,10 +159,7 @@ func WithServer(a p9.Attacher) Set {
 	}
 }
 
-// With9P enables the 9P2000 server in cpu.
-// The server is by default disabled. Ninep is sticky; if set by,
-// e.g., WithNameSpace, the Principle of Least Confusion argues
-// that it should remain set. Hence, we || it with its current value.
+// With9P enables the 9P2000 server in cpu. The server is by default disabled.
 func With9P(p9 bool) Set {
 	return func(c *Cmd) error {
 		c.Ninep = p9
@@ -171,8 +168,7 @@ func With9P(p9 bool) Set {
 }
 
 // WithNameSpace sets the namespace to Cmd.There is no default: having some default
-// violates the principle of least surprise for package users. If ns is non-empty
-// the Ninep is forced on.
+// violates the principle of least surprise for package users.
 func WithNameSpace(ns string) Set {
 	return func(c *Cmd) error {
 		c.NameSpace = ns
--- a/client/client.go
+++ b/client/client.go
@@ -124,9 +124,10 @@ func Command(host string, args ...string) *Cmd {
 		col, row = c, r
 	}
 
+	h, u := GetHostUser(host)
 	return &Cmd{
 		Host:     host,
-		HostName: GetHostName(host),
+		HostName: h,
 		Args:     args,
 		Port:     DefaultPort,
 		Timeout:  defaultTimeOut,
@@ -136,7 +137,7 @@ func Command(host string, args ...string) *Cmd {
 		Row:      row,
 		Col:      col,
 		config: ssh.ClientConfig{
-			User:            os.Getenv("USER"),
+			User:            u,
 			HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 		},
 		hasTTY:  hasTTY,
--- a/client/client.go
+++ b/client/client.go
@@ -411,6 +411,7 @@ func (c *Cmd) Dial() error {
 	}
 	if len(c.FSTab) > 0 {
 		c.Env = append(c.Env, "CPU_FSTAB="+c.FSTab)
+		c.Env = append(c.Env, "LC_GLENDA_CPU_FSTAB="+c.FSTab)
 	}
 
 	return nil
--- a/client/client.go
+++ b/client/client.go
@@ -431,7 +431,7 @@ func (c *Cmd) Start() error {
 	}
 	// Set up terminal modes
 	modes := ssh.TerminalModes{
-		ssh.ECHO:          0,     // disable echoing
+		ssh.ECHO:          1,     // do not disable echoing!
 		ssh.TTY_OP_ISPEED: 14400, // input speed = 14.4kbaud
 		ssh.TTY_OP_OSPEED: 14400, // output speed = 14.4kbaud
 	}
--- a/client/fns.go
+++ b/client/fns.go
@@ -84,7 +84,7 @@ func (c *Cmd) UserKeyConfig() error {
 	}
 	key, err := os.ReadFile(kf)
 	if err != nil {
-		return fmt.Errorf("unable to read private key %q: %v", kf, err)
+		return fmt.Errorf("unable to read private key %q: %w", kf, err)
 	}
 
 	signer, err := ssh.ParsePrivateKey(key)
--- a/client/fns.go
+++ b/client/fns.go
@@ -7,6 +7,7 @@ package client
 import (
 	"bytes"
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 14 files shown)


rminnich:
This may the point at which cpu defaults to nfs instead of 9p.

rminnich:
I wonder ... certainly another way to get the environment intact to the remote program is this:
cpu remote somecommand -env `export` [args ...]
So the remote command would see, for the env switch, the desired environment for the child process. Then, ...
cpu remote cpuns -env `export` [command [switches [args]]]
cpuns would consume the -env just enough to pass it through; and it would process CPU_FSTAB. 

Then the hack with LC_GLENDA_CPU_FSTAB would not be needed at all.

comments?

rminnich:
I am going to redesign this a bit, b/c the ssh restriction on environment variables is truly a nightmare. I think I'll set up if so that cpuns has an arg:
cpuns -env environment-variables-seperated-by-null-chars
where env will be created by the cpu command. 

option to is just coopt LC_TERM and add more names to it. But there's always the change we'll hit more restrictions, as we did with sudo, so that seems a neverending story. 

Look for the rework next week.

rminnich:
I think this is ready for review. Anyone?

client/fns.go
```diff
@@ -111,16 +112,17 @@ func (c *Cmd) HostKeyConfig(hostKeyFile string) error {
...
 func (c *Cmd) SetEnv(envs ...string) error {
+	var err error
 	for _, v := range envs {
 		env := strings.SplitN(v, "=", 2)
 		if len(env) == 1 {
 			env = append(env, "")
 		}
 		if err := c.session.Setenv(env[0], env[1]); err != nil {
-			return fmt.Errorf("Warning: c.session.Setenv(%q, %q): %v", v, os.Getenv(v), err)
+			err = errors.Join(fmt.Errorf("Warning: c.session.Setenv(%q, %q): %v", v, os.Getenv(v), err))
```
binjip978, line 122: maybe it better to add %w instead of %v?

cmds/cpuns/main_linux.go
```diff
@@ -0,0 +1,158 @@
...
+	v("CPUNS:"+f, a...)
+}
+
+func checkprivate() error {
+	mnts, err := mountinfo.GetMounts(mountinfo.SingleEntryFilter("/"))
+	if err != nil {
+		return err
+	}
+	if len(mnts) != 1 {
+		return fmt.Errorf("got more than 1 mount for /(%v):%v", mnts, os.ErrInvalid)
```
binjip978: probably %w in the end

session/session.go
```diff
@@ -165,13 +165,15 @@ func (s *Session) Run() error {
...
 	// This is disappointing to say the least. We just bind a few things into /
 	// bind *may* hide local resources but for now it's the least worst option.
-	if tab, ok := os.LookupEnv("CPU_FSTAB"); ok {
-		verbose("Mounting %q", tab)
-		if err := mount.Mount(tab); err != nil {
-			verbose("fstab mount failure: %v", err)
-			// Should we die if the mounts fail? For now, we think not;
-			// the user may be able to debug. Just record that it failed.
-			s.fail = true
+	for _, fstab := range []string{"CPU_FSTAB"} {
```
binjip978, line 168: do we need looping for one element or something is missing?

session/session.go
```diff
@@ -237,3 +239,60 @@ func New(port9p, cmd string, args ...string) *Session {
...
+	if err := s.TmpMounts(); err != nil {
+		verbose("TmpMounts error: %v", err)
+		s.fail = true
+		errs = errors.Join(errs, err)
+	}
+
+	verbose("call s.NameSpace")
+	err := s.Namespace()
+	if err != nil {
+		return errors.Join(errs, fmt.Errorf("CPUD:Namespace: %v", err))
```
binjip978, line 267: maybe %w?

client/client.go
```diff
@@ -411,6 +408,7 @@ func (c *Cmd) Dial() error {
 	}
 	if len(c.FSTab) > 0 {
 		c.Env = append(c.Env, "CPU_FSTAB="+c.FSTab)
+		c.Env = append(c.Env, "LC_GLENDA_CPU_FSTAB="+c.FSTab)
```
binjip978, line 411: I'm not sure about this, but maybe make it a constant? It's a big string but it used in two different packages and sometimes we need `=` so it maybe a bit awkward...

rminnich: I kinda went back and forth on that, but maybe not just yet, I'm still hoping there is a better way to do this?

client/fns.go
rminnich: done.

session/session.go
rminnich: I think I'll leave it as a loop if that is OK? At some point, we needed it, and might again?

session/session.go
rminnich: done

client/client.go
binjip978: I think it should be, but it not clear to me either.

binjip978 (approved):

cmds/cpu/cpu.go
```diff
@@ -136,7 +131,23 @@ func getPort(host, port string) string {
 }
 
 func newCPU(host string, args ...string) (retErr error) {
-	// note that 9P is enabled if namespace is not empty OR if ninep is true
+	// When running over ssh, many environment variables
```
rjkroege, line 134: This code for building the port and environment seems twisty and exactly the kind of thing that would benefit from being pulled out into utility functions that have some tests to make sure that the environment is sane? Maybe just add a TODO in this PR.

cmds/cpu/doc.go
```diff
@@ -5,116 +5,122 @@
...
-//           use, e.g., /amd64
-//     -sp string
-//          remote port, default 17010
-//     -srv string
-//           what server to run (default none; use internal)
-//     -timeout9p time.Duration
-//           How long to wait for the server to connect to 9p (default100ms)
+//
+//	-9p bool
+//	      enable the 9p server in the client (default enabled)
```
rjkroege, line 75: Weren't there words above about how that this is no longer the case?

cmds/cpu/doc.go
```diff
@@ -254,4 +263,11 @@
 // If you want to learn more read the factotum paper by Grosse et. al. and the original
 // Plan 9 papers on cpu, but be aware there are many subtle details visible only
 // in code.
+//
+// You can now tentatively talk to standard sshd.
+// Here is an example:
+// ./cpu -nfs=true -9p=false -d -key ~/.ssh/homemac -sp 22 127.0.0.1
+// A few things: 9p has to be off, since sshd don't understand the -9 switch
```
rjkroege, line 270: nit: "sshd doesn't understand"

cmds/cpu/doc.go
```diff
@@ -254,4 +263,11 @@
...
+// Failing to Setenv is no longer an error, so you may not be able to set things
+// like SHELL.
```
rjkroege, line 272: It might be helpful to newbies if you note that (at least as I understand the code) requires that the remote system have `sudo` installed?

rjkroege (approved):
Some thoughts added but you can probably ignore me. 😊

cmds/cpu/doc.go
rminnich: fixed. Thanks. Good catch.

PR #280, opened 2024-07-12
client: add GetHostUser for u@host and .ssh/config

rminnich:
Since 2017 cpu used $USER as the user. Now that it supports ssh, it needs better processing.

Add GetHostUser, which takes a string in [user@]host form, and returns a host and user name as follows:
if user@host form, return the host and user.

Look up the host in .ssh/config; if found, and
there is a UserName record, return that UserName

Otherwise return host and $USER.

If a host entry is found in .ssh/config, the
HostName from that entry is used.

This means that code that used GetHost should start using GetHostUser.


client/fns.go
```diff
@@ -200,6 +200,30 @@ func GetHostName(host string) string {
...
+// $USER. Given that just about anything can be valid, and errors
+// will be caught in other places, it does not return an error.
+// Also, experimentally, in the name@host form, name overrides
+// any use name that might be in .ssh/config. Host, on the other hand,
+// is always determined from .ssh/config. The host name
+// is also pulled from .ssh/config; calls to GetHostName
+// can be replaced as needed, but that's for the future.
+func GetHostUser(n string) (host, user string) {
+	if u, h, ok := strings.Cut(n, "@"); ok {
+		host, user = GetHostName(h), u
```
binjip978: You can return right away and make this function for 3 lines shorter.
```go
		return GetHostName(h), u
```
than 
```go
		host, user = GetHostName(h), u
 		return
```

rminnich: yep, thanks, was not thinking.

binjip978 (approved):

PR #281, opened 2024-09-20
Make cpud build on Darwin

rjkroege:
For the convenience of continuous builders, make cpud build (but
not work correctly) on Darwin

Note that this doesn't make cpud _work_ on Darwin -- just compile
successfully.


rminnich (approved):

rminnich:
ah you need a signed-off-by

server/server_other.go
```diff
@@ -14,6 +14,11 @@ import (
 	"os/exec"
 )
 
+type Session struct {
+	binds string
```
rminnich, line 18: remove binds, we can't do that on Darwin IIUC?

rminnich:
I think this is now obsolete? If not, please bring it back, but fix conflicts. Use the freebsd port as a model.

PR #286, opened 2024-10-03
Rust revival

orangecms:
WIP || RIP

lots of work today, not much progress


rminnich:
it's harder than it seems. I think the best bet is write cpuns in rust. Let's close this 2yo thing for now.

PR #289, opened 2024-10-10
Freebsdcpud

rminnich:


cmds/cpud/init_freebsd.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+			if err != nil {
+				log.Printf("CPUD: a process exited with %v, status %v, rusage %v, err %v", p, s, r, err)
+			}
+			numReaped++
+		}
+	}()
+
+	runtime.UnlockOSThread()
+	return nil
```
binjip978, line 80: I don't see how this function can return an error. But is this more or less duplicate, I think it's fine.

cmds/cpud/main_freebsd.go
```diff
@@ -0,0 +1,131 @@
...
+		port9p = flag.String("port9p", "", "port9p # on remote machine for 9p mount")
+
+		flag.Parse()
+		if *debug {
+			v = log.Printf
+			session.SetVerbose(verbose)
+		}
+		// If we are here, no matter what they may set, *remote must be true.
+		// sadly, cpud -d -remote=true -remote=false ... works.
+		*remote = true
```
binjip978, line 88: no sure if I get this, but for me line 69 defines remote value, maybe we should not parse it at all? I mean remove `remote = flag.Bool("remote", false, "indicates we are the remote side of the cpu session")`

binjip978 (approved):

cmds/cpud/main_freebsd.go
rminnich: If we can ever run as init on freebsd, this will matter, so I'd rather leave it in. It follows the pattern of the linux side.

PR #292, opened 2024-12-12
Fix the TESTCPU script and .gitignore for the new world

rminnich:
goanwyhere is really nice, use that.


rminnich:
note that goanywhere let me pass options to the u-root command as before; quite nice

orangecms (approved):

PR #296, opened 2024-12-18
WIP: cpud: initial darwin support

ericvh:
Destigmatize darwin cpud by enabling it
without filesystem mounts or
private namespace.

WIP: need to add nfs mounts and chroot in order for this to really be cpu and not just an ssh clone 

### DIFF
--- a/server/server_other.go
+++ b/server/server_other.go
@@ -8,56 +8,48 @@
 package server
 
 import (
-	"fmt"
-	"net"
 	"os"
 	"os/exec"
 )
 
-// Namespace assembles a NameSpace for this cpud, iff CPU_NAMESPACE
-// is set.
-// CPU_NAMESPACE can be the empty string.
-// It also requires that CPU_NONCE exist.
-func (s *Session) Namespace() (error, error) {
-	var warning error
-	// Get the nonce and remove it from the environment.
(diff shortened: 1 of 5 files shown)


session/session_other.go
```diff
@@ -6,17 +6,13 @@
...
-	"os"
-)
-
 // Namespace assembles a NameSpace for this cpud, iff CPU_NAMESPACE
 // is set.
 // CPU_NAMESPACE can be the empty string.
 // It also requires that CPU_NONCE exist.
 func (s *Session) Namespace() error {
-	return fmt.Errorf("CPUD: 9p mounts are only valid on Linux:%w", os.ErrNotExist)
+	//return fmt.Errorf("CPUD: 9p mounts are only valid on Linux:%w", os.ErrNotExist)
```
rminnich, line 14: I'm unclear on what the reason for this is. Why is it not an error? Why is the error commented out, if nil is OK?

server/server_other.go
```diff
@@ -8,56 +8,48 @@
 package server
 
 import (
-	"fmt"
```
rminnich, line 11: I'm unsure of the point of this file? Is it for Darwin? Will it fool people into thinking things work? Just trying to understand.

PR #298, opened 2025-03-26
Add sysvinit file for milkv-duo S

rminnich:


brho (approved):

orangecms (approved):

orangecms:
The automerge label is obsolete and no longer works.
Instead, GitHub now offers it as an option when creating/changing the PR, and the option needs to be enabled in the settings, which we have here. There is a dropdown within the merge button now that lets you enable the auto-merge for each PR.

PR #304, opened 2025-04-06
CI: upgrade cosign to current version

orangecms:
The older installer was still pulling from googleapis.com, which is
no longer available to the public. See also:
https://github.com/helm/chart-testing-action/issues/132

Signed-off-by: Daniel Maslowski <info@orangecms.org>


rminnich (approved):
Thanks, this is a big fix.

PR #316, opened 2025-10-21
Add TestIP that uses cpu on 4 Linux architectures

rminnich:
The test runs on Linux on amd64, arm, arm64, and riscv64.

Once this is merged, this test can replace the one in u-root. Our goal is to elminate as many u-root VM shell script tests as possible, because we have found undetected errors in them. Tests should be written in Go.

We keep this test here to also test cpu. Getting it all working was no small matter. We want it to keep working :-)

We extend the client package in small ways to make writing VM tests convenient.

The kernel images and initramfs for cpu are built into the package. The tests can be run hermetically by running go test.
ghcr.io is only used for go generate, and that will not be frequent (possibly changing only every year or so).

The kernels are cached from the u-root ghcr.io docker images. The kernels have a few tweaks, for things missing
from the kernel (GRE, VTI) in ghcr.io. The PR for these additions has been submitted for the vmtest repo. The kernels also provide /proc/config.gz. For now, my fixes are local to config_linux.txt files.

Note: I have no idea how the vmscript in u-root is passing at this point, with support for GRE, VTI, and other things missing. The script may not be working; there may be errors that are not caught. I've seen ignored errors in other scripts.

The compiled-in initramfs package only contains cpud. The vm package builds a u-root image for testing as needed and concatenates it to the cpud initramfs.

The VMs start with two network interfaces. The first is used to support cpud; the second interface can be manipulated by the tests. This removes the need to parse VM console output, which can be fragile.

IP is configured by ip=dhcp.

The test assumes there is a working qemu on the CI. This allows us to run these tests on any kernel/architecture combination that has qemu.


vm/vm_linux_test.go
```diff
@@ -59,7 +58,7 @@ func TestCPUAMD64(t *testing.T) {
 		{cmd: "/bbin/dd", args: []string{"if=/tmp/cpu/a", "of=/tmp/cpu/b"}, ok: true},
 	} {
 		cpu, err := i.CPUCommand(tt.cmd, tt.args...)
-		if err != nil {
+		if !errors.Is(err, nil) {
```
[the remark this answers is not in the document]
rminnich: This is incorrect. We just learned the hard way again.

vm/ip_linux_test.go
```diff
@@ -0,0 +1,241 @@
+// Copyright 2025 the u-root Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build !race && amd64
```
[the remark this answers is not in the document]
rminnich: For now, we're limiting to running qemu on amd64.

vm/ip_linux_test.go
```diff
@@ -0,0 +1,241 @@
...
+				{cmd: "ip tunnel del gre_tunnel"},
+				{cmd: "cat /proc/net/dev", excludes: []string{"gre_tunnel"}},
+			},
+			{
+				{cmd: "ip tunnel add vti_tunnel mode vti remote 192.168.242.3 local 192.168.241.1 key 5678", failok: false},
+
+				// Verify VTI tunnel exists in /proc/net/dev
+				{cmd: "cat /proc/net/dev", includes: []string{"vti_tunnel"}},
+
+				//Verify VTI tunnel parameters
```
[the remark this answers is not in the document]
rminnich: fixed. Thanks.

vm/ip_linux_test.go
```diff
@@ -0,0 +1,241 @@
...
+			},
+			{
+				// ARP tests
+				{cmd: "ip neigh add 192.168.241.2 lladdr 00:11:22:33:44:55 dev eth1"},
+				{cmd: "cat /proc/net/arp", includes: []string{"192.168.241.2"}},
+
+				// Verify the neighbor entry
+				{cmd: "ip neigh show dev eth1", includes: []string{"192.168.241.2", "192.168.241.2 dev eth1 lladdr 00:11:22:33:44:55 PERMANENT"}},
+				//{cmd: "test "$neigh_entry" = "192.168.241.2 dev eth1 lladdr 00:11:22:33:44:55 PERMANENT"", includes: []string{},},
+
```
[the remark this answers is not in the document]
rminnich: done

vm/vm_linux_test.go
```diff
@@ -103,9 +103,68 @@ func TestCPUAMD64(t *testing.T) {
...
 	}
+}
+
+// TestCPUARM tests both general and specific things. The specific parts are the io and cmos commands.
+// It being cheaper to use a single generated initramfs, we use the full u-root for several tests.
+func TestCPUARM(t *testing.T) {
+	d := t.TempDir()
+	i, err := vm.New("linux", "arm")
+	if !errors.Is(err, nil) {
+		t.Fatalf("Testing kernel=linux arch=amd64: got %v, want nil", err)
```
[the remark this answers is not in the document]
rminnich: good catch.

brho (approved):

vm/build.go
```diff
@@ -130,12 +130,16 @@ func main() {
 		if len(b.container) == 0 {
 			continue
 		}
-		ref, err := name.ParseReference(b.container)
-		if err != nil {
-			log.Fatal(err)
-		}
+		if false {
```
brho, line 133: remove or flag?  ( i see a bunch of these - if you turn them all on at once, might be easier to do via a single flag)

vm/build.go
```diff
@@ -130,12 +130,16 @@ func main() {
...
+			ref, err := name.ParseReference(b.container, name.Insecure)
+			if err != nil {
+				log.Fatal(err)
+			}
 
-		img, err := crane.Pull(ref.Name())
+			fmt.Printf("parse %s to %s", b.container, ref.Name())
+		}
+		//img, err := crane.Pull(ref.Name())
```
brho, line 141: remove or elaborate?

vm/initramfs_linux_amd64.cpio
brho, line 1: do you want both the cpio and the cpio.gz?

vm/vm_linux_test.go
```diff
@@ -103,9 +103,68 @@ func TestCPUAMD64(t *testing.T) {
 		}
 		t.Logf("io %s = %q", tt.args, string(b))
 	}
+}
+
+// TestCPUARM tests both general and specific things. The specific parts are the io and cmos commands.
+// It being cheaper to use a single generated initramfs, we use the full u-root for several tests.
+func TestCPUARM(t *testing.T) {
+	d := t.TempDir()
```
brho, line 111: future work - would it be convenient to wrap all of the tempdir, vm.New, i.Uroot, start VM, etc. all in a function?

brho (approved):
minor nits

rminnich:
Sorry I missed your comments. I will work on them next.

vm/build.go
rminnich: done

vm/build.go
rminnich: remove.

vm/vm_linux_test.go
rminnich: I think so. I think I want to put some more miles on this to be sure it's the right model. But it's a good suggestion.

PR #318, opened 2025-10-22
remove errors.Is in tests

rminnich:
errors.Is(err, nil) is needed, in some places, but not here.

Sometimes packages return error wrapping nil, but not in this case.

But, see:
https://go.dev/play/p/yjBlsllddSE

Gemini confidently predicts that the code in the Playground is wrong. Oops. Thanks to brho for that test.


brho (approved):
errors that wrap nil are the devil

